PR #4, opened 2015-09-28
changes made to the On-Boarding Excercise.

augustineanosike:
Please review my code


JeremyWalters:
Hi,

Can you please squash you commits so it can only be one, secondly there is a lot of files part of your commit that shouldn't be there. We only want to see the code changes you have worked on. Add the other to the ignore list.

Thanks

PR #10, opened 2015-10-07
Onboarding for Roan Brand

RoanBrand:
3rd try

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,8 @@
+bin/*
+obj/*
+packages/*
+Scripts/*
+app.js
+app.js.map
+packages.config
+*.suo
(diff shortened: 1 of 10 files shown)


JacoBriers:
Some general comments on the 6 points mentioned in the brief:
1. Brief correctly followed, grid works as expected. Thumbs up here
2. Code Style: Generally good, just need to use namespaces to avoid polluting the global namespace
3. User-centric thinking: Pretty easy to use, Some small issues like repeating the bottom element again and not having a way to jump to pages toward end of list (what if I wanted to scroll to page 500?). Loading indication would have been nice touch
4. Suitable comments : Good stuff
5. Performance : good, only fetched data required, cached column names, but re-fetched record-count with every page. 
6. Aesthetics : might need a bit of help from Elize and Thea :)

PR #20, opened 2016-03-11
Bradley - Onboarding

bradleymeintjes:
Submitting for review


src/rowfactory.ts
```diff
@@ -0,0 +1,40 @@
...
+
+		/*
+		* Generator class for new table rows/cells
+		*/
+		export class RowFactory {
+
+			/*
+			* Creates a new table row element populated with table header cells with the given data
+			*/
+			createHeaderRow(values: Array<string>): JQuery {
```
JacoBriers, line 12: We prefer `string[]` rather than `Array<string>` -> less typing

src/grid.ts
```diff
@@ -0,0 +1,146 @@
+﻿/// <reference path="../scripts/typings/jquery/jquery.d.ts" />
```
JacoBriers, line 1: If you include a file in the Visual Studio solution, there is no need to add this reference

src/app.ts
```diff
@@ -0,0 +1,95 @@
+﻿let grid: iq.onboard.Grid,
+	refreshHandle: number = 99;
+
+$(() => {
+	grid = new iq.onboard.Grid("#grid");
+
+	initialize();
+});
+
+function initialize() {
```
JacoBriers, line 10: If you define a function outside a module it will pollute the global namespace, and can cause some issues if someone else uses the same naming (third party libraries are the biggest risk)

src/app.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function initControlPanel() {
+	initControlPanelAnimation();
+	initNavigationControlAnimation();
+	initPageSelector();
+}
+
+function initControlPanelAnimation() {
+	let $controlPanel = $(".control-panel");
```
JacoBriers, line 52: Using a CSS class selector can return multiple JQuery instances. If you had another DOM element re-using the same CSS class (which is likely because re-usability is the purpose of CSS classes), your event handler would be attached to both elements. 

We prefer using ID selectors when attaching event handlers unless you are absolutely sure that the CSS class won't be re-used elsewhere (but I don't ever take that chance)

scripts/onboard.js
```diff
@@ -0,0 +1,260 @@
+var grid, refreshHandle = 99;
```
JacoBriers, line 1: Because this file gets generated by the TS compiler, there is no need to commit it

src/grid.ts
```diff
@@ -0,0 +1,146 @@
...
+				if (page >= this.numberOfPages) {
+					console.log("Cant page to " + (page+1) + " because there are only " + this.numberOfPages + " pages!");
+					page = this.numberOfPages-1;
+				}
+				if (page < 0) {
+					console.log("Page number must be positive!");
+					page = 0;
+				}
+
+				console.log("Page: " + page);
```
JacoBriers, line 60: This is fine for debugging, but for some versions of IE the console is disabled if the browser is not in developer mode and calling `console.log` would cause an exception. We have a wrapper in the IMQS code, `iq.log`, that checks if `window.console` is defined before using the function. You need to do something similar if you want to be safe.

PR #21, opened 2016-06-08
Danie onboard-javascript

dbferreira:
Hi guys, please review...


app.css
```diff
@@ -0,0 +1,38 @@
...
+.paging div {
+    float: left;
+    margin-right: 5px;
+}
+
+#data-table {
+    clear: both;
+    padding-top: 10px;
+    width: 100%;    
+    white-space: nowrap
```
ppottas: semicolon

app.ts
```diff
@@ -0,0 +1,113 @@
+﻿let totalRecords = 0;
+let indexPos = 0;
+
+class AjaxCall {
+	url: string = "http://localhost:2050";
```
ppottas: make applicable members private

app.ts
```diff
@@ -0,0 +1,113 @@
...
+	indexFinish: number;
+	fullQuery: string;
+	request: string;
+	objHttpReq: any;
+
+	doAjaxCall(request: string, callback: Function, start?: number, finish?: number) {
+		this.request = request;
+		this.indexStart = start;
+		this.indexFinish = finish;
+		if (start >= 0) { this.fullQuery = this.url + "/" + request + "?from=" + start + "&to=" + finish; }
```
ppottas: `start` may be `undefined`

ppottas: See our [JavaScript Style Guide](https://imqssoftware.atlassian.net/wiki/display/AR/JavaScript+Style+Guide). We normally put our statements on a new line after `{`.

app.ts
```diff
@@ -0,0 +1,113 @@
...
+		else { this.fullQuery = this.url + "/" + request; }
+		this.objHttpReq = new XMLHttpRequest();
+		this.objHttpReq.onreadystatechange = () => this.onRStateChange(callback);
+		this.objHttpReq.open("Get", this.fullQuery, true);
+		this.objHttpReq.send();
+	}
+
+	onRStateChange(callback: Function) {
```
ppottas: make private

ppottas: Instead of `Function` consider using `(your parameters) => void`

app.ts
```diff
@@ -0,0 +1,113 @@
...
+			let data = JSON.parse(this.objHttpReq.responseText);
+			(callback)(data, this.indexStart, this.indexFinish);
+		}
+		// Handle error responses here if required
+	}
+}
+
+// Create table header row based on columns received
+function populateColumns(columns: JSON) {
+	let columnArray = Object.keys(columns).map(function (k) { return columns[k]; });
```
ppottas: Use closures: `(k) => { return columns[k]; }`

app.ts
```diff
@@ -0,0 +1,113 @@
...
+		th.innerHTML = columnArray[i];
+		tr.appendChild(th);
+	}
+	htmlTable.appendChild(tr);
+}
+
+// Total record counter update.  Only called once on page load
+function updateRecordCounter(recordCount: JSON) {
+	totalRecords = Number(JSON.stringify(recordCount));
+	document.getElementById("page-total").innerHTML = String(totalRecords);
```
ppottas: you can simply `...innerHTML = recordCount.toString();`

app.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Update table with content received from API call
+function updateGrid(records: JSON, start: number, finish: number) {
+	indexPos = start;
+	document.getElementById("page-progress").innerHTML = "Displaying " + start + " to " + finish + " of ";
```
RoanBrand: You can use template strings  `Displaying ${start} to ${finish} of  `;

app.ts
```diff
@@ -0,0 +1,113 @@
...
+	let recordArray = Object.keys(records).map(function (k) { return records[k]; });
+	let tableContent = document.getElementById("table-body");
+	// Clear old table content
+	while (tableContent.firstChild) {
```
RoanBrand, line 66: There should be a function to clear this in one go. I see this method is supposedly faster: http://stackoverflow.com/questions/3955229/remove-all-child-elements-of-a-dom-node-in-javascript

In IMQS we will probably prefer jquery's `.empty()`

app.ts
RoanBrand: Template string also prefered here. Better when building string with more than 2 +'s

app.ts
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+window.onload = () => {
+	// Get brower height and calculate number of rows to display
+	let browserHeight = window.innerHeight - 205;
+	let pageSize = Math.floor(browserHeight / 33);
+
+	// Initialise Ajax calls and retrieve initial data
+	let ajColumns = new AjaxCall();
```
RoanBrand: use const as far as possible. only when the reference to a value is changed for an identifier is let required.

app.ts
```diff
@@ -0,0 +1,113 @@
...
+	};
+	buttonPrevious.onclick = () => {
+		let previousStart = indexPos - pageSize - 1;
+		if (previousStart < 0) ajRecords.doAjaxCall("records", updateGrid, 0, pageSize);
+		else ajRecords.doAjaxCall("records", updateGrid, previousStart, indexPos - 1);
+	};
+	buttonNext.onclick = () => {
+		let nextFinish = indexPos + pageSize * 2 + 1;
+		if (nextFinish >= totalRecords) ajRecords.doAjaxCall("records", updateGrid, totalRecords - pageSize - 1, totalRecords - 1);
+		else ajRecords.doAjaxCall("records", updateGrid, indexPos + pageSize + 1, nextFinish);
```
RoanBrand: newline after conditional and else

app.ts
ppottas, line 66: If you remove table-body and create it again as well as add all records internally, then finally updating the DOM once, you'd gain a slight performance increase, but the increase would only be noticeable (and true) when adding 100's of thousands of records.

app.ts
```diff
@@ -22,11 +22,15 @@ class AjaxCall {
...
-		if (this.objHttpReq.readyState === 4 && this.objHttpReq.status === 200) {
-			let data = JSON.parse(this.objHttpReq.responseText);
-			(callback)(data, this.indexStart, this.indexFinish);
+		if (this.objHttpReq.readyState === XMLHttpRequest.DONE) {
+			if (this.objHttpReq.status === 200) {
+				let data = JSON.parse(this.objHttpReq.responseText);
+				(callback)(data, this.indexStart, this.indexFinish);
+			}
+			else {
+				console.log("AJAX error with the following status occurred: " + this.objHttpReq.status);
```
JeremyWalters, line 33: We never you console.log. It just causes problems and populate console with useless information. We dont even use it for debugging because one tend to forget to remove it when done.

JacoBriers, line 33: Wow harsh... ;) Jeremy is correct that we use [little wrappers functions](https://github.com/IMQS/www/blob/master/js/js-base/defs.ts#L113-L178) for `console.log` inside the www codebase: `iq.logErr` and its friends `iq.logWarn`, `iq.logInfo`, `iq.logDebug` and `iq.logTrace`. For the onboarding project, however, you don't have access to those functions, so using `console.log` in this case is completely fine. 

The reason we don't use it directly in our production codebase is because if the IT department at a client deactivated the developer tools in IE, `console.log` would cause an error. 

On the amount of info you log: I would say feel free to spam away, more info is better than none, just make sure to use the correct log level.

PR #24, opened 2017-04-10
Michael-Onboard-Typescript

ghost:


RoanBrand:
Need to squash commits

RoanBrand:
Also make a `.gitignore`, excluding all the IDE generated files, binary files, etc.

Grid/app.ts
```diff
@@ -0,0 +1,28 @@
+﻿class Greeter {
+    element: HTMLElement;
```
dbferreira: Changes spaces to tabs

Grid/app.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    stop() {
+        clearTimeout(this.timerToken);
+    }
+
+}
+
+window.onload = () => {
+    var el = document.getElementById('content');
+    var greeter = new Greeter(el);
```
dbferreira: Rather use `const` in place of `var`

app.ts
```diff
@@ -0,0 +1,106 @@
...
+		}
+	});
+
+	// Retrieves the max amount of records from the API GET call
+
+	function getMax() {
+		$.getJSON("/recordCount", function (data) {
+			totRec = data;
+		});
+		return totRec;
```
dbferreira: This function will start the async `getJSON` call, and then immediately return `totRec`, which would still be undefined.  Move the return statement inside the callback function of the `getJSON` call to return after the result has been fetched.

It probably still works when you've tested it because you're using global variables and the value is set regardless of the return statement.

app.ts
```diff
@@ -0,0 +1,106 @@
...
+namespace onboard {
+
+	// Variable declarations
+
+	let totRec = getMax(); // Total amount of records
+	let totHeadings;
+	let fromRecord = 0;
+	let toRecord = 15;
+	let count = 0;
+	let increase = 15;
```
dbferreira: For these variables, define a type, e.g. `const totRec: number = getMax();`.  You can use const for the values that would not change.

app.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	$(document).ready(function () {
+
+		// Creation of table as well as insertion of data
+
+		function populate(to, from) {
```
dbferreira: Define types for the input variables, and also define a return type for the function.

app.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Michael Leibbrandt 
+ * Typescript Onboard Project
+ */
+
+namespace onboard {
+
+	// Variable declarations
+
```
dbferreira: It would be good to make use of a `class` with a constructor here.

app.ts
```diff
@@ -0,0 +1,106 @@
...
+				} else if (count === totRec) {
+					$('#btn1').prop("disabled", true);
+				}
+				// Retrieves and sets column names
+
+				let headers = "<tr>";
+				for (let i = 0; i < data.length; i++) {
+					headers += `<th>${data[i]}</th>`
+				}
+				totHeadings = data.length;
```
dbferreira: If you move this declaration up before your `for` loop you can use the same variable (`totHeadings`) for instantiating the for-loop.

app.ts
```diff
@@ -0,0 +1,106 @@
...
+				headers += "</tr>";
+				$("#grid").append(headers);
+			});
+
+			let records = "<tr>";
+			if (to <= from) to = from;
+			$.getJSON("/records?from=" + from + "&to=" + to, function (data) {
```
dbferreira: Change the `function` to an array function

app.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	$(document).ready(function () {
+
+		// Creation of table as well as insertion of data
+
+		function populate(to, from) {
+
+			// Retrieves columns or Headings from API GET call
+
+			$.getJSON("/columns", function (data) {
```
dbferreira: Use array functions

*edit: arrow functions

app.ts
```diff
@@ -0,0 +1,106 @@
...
+
+		function previous() {
+			$('#btn1').prop("disabled", false);
+			$('#grid td').remove();
+			$('#grid tr').remove();
+			populate((toRecord - increase), (fromRecord - increase));
+			toRecord -= increase;
+			fromRecord -= increase;
+			count -= 1;
+			console.log(count);
```
dbferreira: We usually remove all `console.log` calls that might have been used during development.  If you still want to log something you can use one of our `iq.log...` functions.

dbferreira (changes requested):
You're on the right track.  My biggest suggestion would be to use a class for the main app.ts file. 

I have not seen any logic for resizing the content when scaling the window.

app.css
```diff
@@ -0,0 +1,26 @@
+﻿html {
+	height: 100%;
+}
+
+html, body {
+	
+}
```
RoanBrand: Why this empty selector?

app.ts
```diff
@@ -0,0 +1,95 @@
...
+namespace onboard {
+
+	let totRec: number;
+	let totHeadings: number;
+	let fromRecord: number = 0;
+	let increase: number;
+	let resizeCounter;
+
+
+		$.getJSON("/recordCount", (data) => {
```
RoanBrand: indentation

app.ts
```diff
@@ -0,0 +1,95 @@
...
+			totRec = data;			
+		});
+		winResize();
+	
+	
+	// Makes an onresize trigger event
+	window.onresize = winResize;
+
+	// Binds button actions once window is ready
+	$(window).ready(() => {
```
RoanBrand: can shorten to `$(() => {`

app.ts
```diff
@@ -0,0 +1,95 @@
...
+	// Cleans up and retrieves next table of data		
+	function next() {
+
+		fromRecord += increase;
+		$('#prvButton').prop("disabled", false);
+		populate(fromRecord);
+	}
+
+	function populate(from) {
+		console.log("Populated!");
```
RoanBrand: remove `console.log`s when you are done

RoanBrand (changes requested):
The grid does not not cover the complete area when window is maximized. And when I click the next/previous buttons too fast a few times the table breaks. It sometimes duplicates the header in the middle or bottom of the grid:
image

PR #25, opened 2017-04-20
Legend ajs onboarding

FrancoRootman:


.gitignore
```diff
@@ -0,0 +1,2 @@
+*.js
+*.map
```
navonod, line 2: Also add your IDE files to be ignored

app.ts
```diff
@@ -0,0 +1,163 @@
...
+            this.$row.append(this.$header.clone().text(col));
+        }
+
+        $("#grid").append(this.$row.clone());
+    }
+
+    //iterates through records and adds them as rows to the grid
+    drawGrid(records: string[][]) {
+        for (let record of records) {
+            this.$row.empty();
```
navonod: I _think_ this is an avoidable expense. The clone() that you use later makes a new object instantiation, which you could just as well do from a method that returns the JQuery result you have for the row member creation above. You wouldn't need the row member then, and the need to empty it each time would be negated.

app.css
```diff
@@ -0,0 +1,48 @@
...
+#btns {
+    z-index:9999;
+    margin-left:4px;
+    margin-top: 391px;
+    position:fixed;
+    height:170px;
+    width:50px;
+}
+
+#prevbtn, #nextbtn {
```
bradleymeintjes, line 44: For the sake of re-usability and maintainability, this is a good situation to use a class instead of ids. I would say most css styling should target classes rather than ids

app.ts
```diff
@@ -0,0 +1,163 @@
...
+                //adds initial records to row objects to be added to grid on page load.
+                $.getJSON(host + "/records?from=" + this.from + "&to=" + this.to, (data) => { })
+                    .done((records) => {
+                        this.drawGrid(records);
+                    });
+            });
+    }
+
+    //(re)initializes the page deafault values
+    pageInit() {
```
navonod: This could just become a clearing of the main div, and the redrawing of the header.

app.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        $("#grid").append(this.$row.clone());
+    }
+
+    //iterates through records and adds them as rows to the grid
+    drawGrid(records: string[][]) {
+        for (let record of records) {
+            this.$row.empty();
+
+            for (let col of record) {
```
navonod: You can extract this logic into a new method called drawRow(). Because the header and a row are identical, you can use drawRow to create the header row too - just pass the css class as a method param.

app.css
```diff
@@ -0,0 +1,48 @@
...
+    outline: 1px solid;
+    display: inline-block;
+    background: #FFFFFF;
+}
+
+.header {
+    background-color: lightsteelblue;
+}
+
+#btns {
```
bradleymeintjes: This is not descriptive enough. Since it is a container for the prev/next, we should probably name it accordingly. Also refer to the comment about using classes below

app.ts
```diff
@@ -0,0 +1,163 @@
+﻿const host = "http://localhost:2050";
+let grid: Grid;
+const initRecord = 0;
+const rowLimit = 22;
+
+class Grid {
+    from: number;
```
bradleymeintjes: variables are public by default if im not mistaken, maybe make these private

app.ts
```diff
@@ -0,0 +1,163 @@
...
+    to: number;
+    headers: string[] = [];
+
+    $row = $("<div />", {
```
bradleymeintjes: I like the $var convention, just know that in our code we dont really use it.

app.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    $header = $("<div />", {
+        class: 'header'
+    });
+
+    $column = $("<div />", {
+        class: 'column'
+    });
+
+    //sets initial grid values.
```
bradleymeintjes: We have a ts linter set up on our front end. One of the rules is that comments must have a space between the // and the first character

navonod (commented):
Just a few suggestions.

app.ts
```diff
@@ -0,0 +1,163 @@
...
+                    });
+            });
+    }
+}
+
+$(document).ready(function () {
+    grid = new Grid(initRecord, initRecord + rowLimit);
+
+    $("#prevbtn").click(
+        () => {
```
bradleymeintjes: You can just pass grid.previousPage() as the parameter here

app.ts
```diff
@@ -0,0 +1,163 @@
...
+    $column = $("<div />", {
+        class: 'column'
+    });
+
+    //sets initial grid values.
+    constructor(from: number, to: number) {
+        this.from = from;
+        this.to = to;
+
+        $.getJSON(host + "/columns", (data) => { })
```
bradleymeintjes: I suppose the host should also be passed into this class to decouple it from the global variables

app.ts
```diff
@@ -0,0 +1,163 @@
...
+            .done((headers) => {
+                this.headers = headers;
+                this.pageInit();
+
+                //adds initial records to row objects to be added to grid on page load.
+                $.getJSON(host + "/records?from=" + this.from + "&to=" + this.to, (data) => { })
```
bradleymeintjes: This bit is repeated, rather move it out into a method so we have one instance of it to maintain

app.ts
bradleymeintjes: most of the methods can be private, also in our production code we define a return type for methods, even in a case like this it would have a : void return type

app.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    //(re)initializes the page deafault values
+    pageInit() {
+        $('#grid').empty();
+        this.$row.empty();
+        this.$column.empty();
+
+        //add headers to row object to be added to the grid on initialization.
+        for (let col of this.headers) {
```
bradleymeintjes: i would move this row creating code into a method (probably createRow) so that it can be reused in drawGrid()

app.ts
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    //loads the higher id value set of records, than the current set of records, to display on the grid.
+    nextPage() {
+        this.pageInit();
+        this.from = this.to + 1;
+        this.to = this.to + rowLimit + 1;
+
+        $.getJSON(host + "/recordCount", (result) => {
```
bradleymeintjes: the urls should be build once and stored in a class variable if used in multiple places.

bradleymeintjes: Also, do you think it is a good idea to have the recordCount checked each time the button is clicked?

app.ts
```diff
@@ -0,0 +1,163 @@
...
+    //loads the record set that starts and ends on opposite sides of the repo reset. ie from = 999982, to = 4
+    private recordRollover(recordCount: number) {
+        let records: string[][] = [];
+        console.log("FROM: " + this.from + " to " + this.to);
+        $.getJSON(host + "/records?from=" + this.from + "&to=" + (recordCount - 1), (res) => { })
+            .done((res) => {
+                records = res;
+            }).then((todo) => {
+                $.getJSON(host + "/records?from=" + 0 + "&to=" + this.to, (res) => {
+                    console.log("GETS HERE");
```
bradleymeintjes: heh

app.ts
FrancoRootman: I wanted to load it to its own var but figured in a real world setting where the recordCount might have changed since the last click, it would be better to check it again. but in this case ...  no its not

app.ts
```diff
@@ -30,110 +22,123 @@ class Grid {
...
+        });
+
+        let row = this.newRow();
 
-            $("#grid").append(this.$row.clone());
+        for (let col of record) {
+            row.append($colu.clone().text(col));
         }
+
+        $("#grid").append(row.clone());
```
navonod: You shouldn't need to clone() the row here, just add the row that's been newly created. Cloning it makes the operation unnecessarily expensive because you're essentially creating two rows when you only need one.

app.ts
```diff
@@ -30,110 +22,123 @@ class Grid {
...
+            this.drawRow("column", record);
+        }
+    }
 
-            for (let col of record) {
-                this.$row.append(this.$column.clone().text(col));
-            }
+    // draws individual rows by appending column objects to it and appending the row object to the grid.
+    private drawRow(css_class: string, record: string[]): void {
+        let $colu = $("<div />", {
```
navonod: I'll grant you that it's debatable as to weather this pattern is necessary, but for consistency's sake I think it would be best that we use a newColumn() method for this, and use it inline as a param to row,append() below - in the same way we've done for newRow()

app.ts
```diff
@@ -30,110 +22,123 @@ class Grid {
...
+            class: css_class
+        });
+
+        let row = this.newRow();
 
-            $("#grid").append(this.$row.clone());
+        for (let col of record) {
```
navonod: You'll find that much of the www code at imqs would does not include curly brackets when dealing with single-line if & for-loop statements.

app.ts
```diff
@@ -30,110 +22,123 @@ class Grid {
...
-                            this.drawGrid(res);
-                        });
-                }
-            });
+        if (this.from < 0 && this.to >= 0) {
+            this.from = this.from + this.recordCount;
+            this.recordRollover(this.recordCount);
+        }
+        else {
+            if (this.from == this.recordCount) {
```
navonod: Yeah, single-line ifs could also give the boot to the curlies.

navonod (commented):
Just a few more comments

PR #26, opened 2017-05-04
Stewart onboard javascript

stewartm83:
Onboard solution can you review please.


FritzOnFire:
Haven't looked at the code yet, but the grid should fill the screen automatically, also when changing the window size we don't want to see a scroll bar, and if I hit the ">>" first the page does not update, and the text changes to "Page NaN of 100000", which is definitely not right.

FritzOnFire:
Please ignore your .js files with .gitignore

app.ts
```diff
@@ -0,0 +1,87 @@
...
+	baseUrl = "http://localhost:2050/";
+
+	constructor(pageSize: number, pageNumber: number) {
+		this.pageNumber = pageNumber;
+		this.pageSize = pageSize;
+	}
+
+	// Load all columns returned from the server
+	private loadColumns(data) {
+		var header = document.createElement("thead");
```
FritzOnFire: don't use `var`

app.ts
```diff
@@ -0,0 +1,87 @@
...
+
+		header.appendChild(headerRow);
+		$("#recordsTable").append(header);
+	}
+
+	// Process and render the data that is coming from the server
+	private loadRecords(data) {
+		// Clear the table body before adding new content
+		$("#recordsTable").children("tbody").html("");
+		var tableBody = document.createElement("tbody");
```
FritzOnFire: don't use `var`

app.ts
```diff
@@ -0,0 +1,87 @@
...
+
+		// Add data items to the row
+		$.each(data,
+			(key, val) => {
+				var tableRow = document.createElement("tr");
```
FritzOnFire: don't use `var`

app.ts
```diff
@@ -0,0 +1,87 @@
...
+				tableRow.innerHTML =
+					`<td>${val[0]}</td><td>${val[1]}</td><td>${val[2]}</td><td>${val[3]}</td><td>${val[4]
+					}</td><td>${val[5]}</td><td>${val[6]}</td><td>${val[7]}</td><td>${val[8]
+					}</td><td>${val[9]}</td><td>${val[10]}</td>`;
```
FritzOnFire: I don't think this would work if you received more or less columns from the server.

FritzOnFire (changes requested):
You should have also used the JQuery that was included with the project, but I don't think that is such a big deal

FritzOnFire:
Also, the first time I click the `>` button, nothing happens

FritzOnFire:
I cant see the navigation buttons when the height of the window is small. And clicking the `>>` button after resizing creates a console error

RoanBrand:
This is hard to review with all the commits. Should he squash them before we give a review? @bmharper @dbferreira

bmharper:
I agree - can you squash your commits @stewartm83? If you've never squashed git commits, you'll probably want to go through some tutorial on doing it.

stewartm83:
Should it be closed?

PR #28, opened 2018-11-20
Onboarding Project - Nkosi

nkosimarz:
Hi all 

Here is my solution to the javascript onboarding project.

Regard
Nkosi


app.ts
```diff
@@ -0,0 +1,140 @@
...
+        let numRows = Math.floor((contentHeight - CONTROLS_HEIGHT - 26) / 46);
+        numRows = numRows <= 0 ? 1 : numRows;
+
+        // set number of records to get from server and ensure that the last record doesnt exceed the limit.
+        let endRecord = (firstRecord + numRows - 1);
+        endRecord = (endRecord >= RECORD_COUNT ? RECORD_COUNT - 1 : endRecord);
+
+        // get data from server
+        const columnsString = getData("http://localhost:2050/columns");
+        let recordsText = getData("http://localhost:2050/records?from=" + firstRecord + "&to=" + endRecord);
```
Winston-Guess: This should be const

app.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        // parse text from api call into arrays
+        const columnsArray = parseText(columnsString);
+        let recordsTable = parseText(recordsText);
```
Winston-Guess: const

app.ts
```diff
@@ -0,0 +1,140 @@
...
+            let th = document.createElement('th');
+            this.myTable.appendChild(th);
+
+            th.appendChild(document.createTextNode(" " + i));
+        }
+
+        // write html for table
+        for (const iEntry of recordsTable) {
+
+            let tr = document.createElement('tr');
```
Winston-Guess: const

app.ts
```diff
@@ -0,0 +1,140 @@
...
+            this.myTable.appendChild(tr);
+
+            for (const jEntry of iEntry) {
+
+                let td = document.createElement('td');
```
Winston-Guess: const

app.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    let newfirstEntry = CURRENT_ENTRY - (LAST_ENTRY - CURRENT_ENTRY) - 1;
+
+    // if the new first entry is less than zero, take user to the first page 
+    newfirstEntry <= 0 ? createPage(0) : createPage(newfirstEntry);
+}
+
+// hint for the user
+function helpFunc() {
+    alert('Table for viewing records with IDs 0 to '
```
Winston-Guess: IMO a better way to do multi-line (or longer) strings is with backticka as follows
```suggestion
    alert(`
        Table for viewing records with IDs 0 to ${RECORD_COUNT - 1}  (Numbers Only)
        To 
```

(Not sure why the 'with' is highlighted in orange 😕

app.ts
```diff
@@ -0,0 +1,140 @@
...
+        + (RECORD_COUNT - 1) + ' (Numbers Only)\nTo start from a specific record'
+        + 'enter a record ID and click "Go To"')
+}
+
+window.onresize = () => {
+    clearTimeout(resizeTimer);
+    resizeTimer = setTimeout(doneResizing, 500);
+
+    function doneResizing() {
```
Winston-Guess: You're doing crazy stuff here, no? Looks like you're abusing javascript 🥇. Also, you should always be using arrow functions these days.

app.ts
```diff
@@ -0,0 +1,140 @@
...
+        createPage(CURRENT_ENTRY);
+    }
+}
+
+// load page for the first time and create buttons
+window.onload = () => {
+
+    let controlsDiv = this.controls;
```
Winston-Guess: should be `const`.

app.ts
```diff
@@ -0,0 +1,140 @@
...
+    CONTROLS_HEIGHT = controlsDiv.offsetHeight;
+
+    // html for buttons and input text fields
+    controlsDiv.innerHTML += '<button onclick = "createPage(0)">First Page</button>';
```
Winston-Guess: You can use backticks here as well. I'm going to leave it to you to figure out how ^^

app.ts
```diff
@@ -0,0 +1,140 @@
...
+
+function createPage(firstRecord) {
+
+    // ensure that the user input is a number and within range of available records
+    if (firstRecord >= RECORD_COUNT || firstRecord < 0 || isNaN(firstRecord)) {
+        alert('please select a NUMBER between 0 and ' + (RECORD_COUNT - 1));
+        return;
+    }
+
+    let contentsDiv = this.content;
```
Winston-Guess: should be `const`.

Read this https://mathiasbynens.be/notes/es6-const

app.ts
```diff
@@ -0,0 +1,140 @@
...
+    contentsDiv.innerHTML = ''; // empty table div to fill in a new table
+    let tableElement = new tableGenerator(contentsDiv, parseInt(firstRecord)); // create table using the tableGenerator class
```
Winston-Guess: Does this get used anywhere?

app.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    // keep the ID of the first and last table entries to use for next and previous page buttons
+    LAST_ENTRY = contentsDiv.lastChild.lastChild.firstChild.textContent;
```
Winston-Guess: This is very obscure. I'd rather get the div by id if you can so that the code is more readable

app.ts
```diff
@@ -0,0 +1,140 @@
...
+
+// fucntion used by the next page buttton, requires the last entry from previously generated table
+function nextPage() {
+    createPage(parseInt(LAST_ENTRY) + 1);
+}
+
+// fucntion used by the previous page buttton, requires the first and last entry from previously generated table
+function prevPage() {
+
+    let newfirstEntry = CURRENT_ENTRY - (LAST_ENTRY - CURRENT_ENTRY) - 1;
```
Winston-Guess: const

app.ts
```diff
@@ -0,0 +1,140 @@
...
+    myTable: HTMLElement;
+
+    constructor(element: HTMLElement, firstRecord) {
+        this.element = element;
+        this.myTable = document.createElement('table');
+        this.element.appendChild(this.myTable);
+
+        // get table height based on  remaining window size after buttons
+
+        let contentHeight = (window.innerHeight - CONTROLS_HEIGHT);
```
Winston-Guess: const

app.ts
```diff
@@ -0,0 +1,140 @@
+﻿// Onboarding Project - 2018
+
+const RECORD_COUNT = JSON.parse(getData("http://localhost:2050/recordCount")); // total number of records available in webserver
+let LAST_ENTRY, CURRENT_ENTRY; // entry numbers required for prev and next funtionality
+let resizeTimer; // delay required to reduce the number of calls made by browser when resizing the window
+let CONTROLS_HEIGHT;
+
+// makes api call and returns data
+function getData(dataLink: string) {
+    let apiCall = new XMLHttpRequest();
```
Winston-Guess: Should be `const`

app.ts
```diff
@@ -0,0 +1,140 @@
...
+    apiCall.open("GET", dataLink, false); // false to prevent asych // warning need to change
```
Winston-Guess: I think you should use an async method here

app.ts
```diff
@@ -0,0 +1,140 @@
...
+    apiCall.send(null);
+    return apiCall.responseText;
+}
+
+// converts strings into array and table
+function parseText(textString) {
+    let outputArray = textString.replace(/'/g, '""');
+    outputArray = JSON.parse(textString);
```
Winston-Guess: This is naughty, I'm going to leave it to you to figure out why ^^

app.ts
```diff
@@ -0,0 +1,140 @@
...
+    return outputArray;
+}
+
+class tableGenerator {
+    element: HTMLElement;
+    myTable: HTMLElement;
+
+    constructor(element: HTMLElement, firstRecord) {
```
Winston-Guess: what is firstRecord?

app.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        let contentHeight = (window.innerHeight - CONTROLS_HEIGHT);
+        this.myTable.setAttribute("height", "contentHeight");
+
+        // calculate row number based on table header + row heights + borders in CSS and ensure its never less than 1
+        let numRows = Math.floor((contentHeight - CONTROLS_HEIGHT - 26) / 46);
+        numRows = numRows <= 0 ? 1 : numRows;
+
+        // set number of records to get from server and ensure that the last record doesnt exceed the limit.
+        let endRecord = (firstRecord + numRows - 1);
```
Winston-Guess: What is first record and you don't need brackets here

Winston-Guess (changes requested):

app.ts
nkosimarz: I thought ' was ` so couldn't get it to work and stuck to ' ' strings

app.ts
nkosimarz: changed it so that table element  has interface as opposed to auto generating the table

tableElement.generateTable();

app.ts
```diff
@@ -0,0 +1,147 @@
...
+
+// load page for the first time and create buttons
+window.onload = () => {
+
+    const controlsDiv = this.controls;
+    CONTROLS_HEIGHT = controlsDiv.offsetHeight;
+
+    // html for buttons and input text fields
+    controlsDiv.innerHTML += `
+    <button onclick = "createPage(0)">First Page</button>
```
Winston-Guess: Better but should be indented

app.ts
```diff
@@ -0,0 +1,147 @@
...
+    const newfirstEntry = CURRENT_ENTRY - (LAST_ENTRY - CURRENT_ENTRY) - 1;
+
+    // if the new first entry is less than zero, take user to the first page 
+    newfirstEntry <= 0 ? createPage(0) : createPage(newfirstEntry);
+}
+
+// hint for the user
+function helpFunc() {
+    alert(`
+    Table for viewing records with IDs 0 to ${(RECORD_COUNT - 1)} (Numbers Only).
```
Winston-Guess: indent here

app.ts
```diff
@@ -0,0 +1,147 @@
+﻿// Onboarding Project - 2018
+
+const RECORD_COUNT: number = JSON.parse(getData("http://localhost:2050/recordCount")); // total number of records available in webserver
+let LAST_ENTRY, CURRENT_ENTRY; // entry numbers required for prev and next funtionality
```
Winston-Guess: There a reason you're using caps here?

Winston-Guess (commented):
A few last things from my side

app.ts
nkosimarz: i think this is already indented

app.js
```diff
@@ -0,0 +1,111 @@
+// Onboarding Project - 2018
+var _this = this;
```
JacoBriers: Just something to keep in mind when you're setting up a project from scratch: don't commit generated files or build artifacts because devs will generate their own when they build the project. Typically these artifacts are added to the `.gitignore` file and won't show up when you run a `git commit`.

app.ts
```diff
@@ -0,0 +1,147 @@
...
+let resizeTimer; // delay required to reduce the number of calls made by browser when resizing the window
+let CONTROLS_HEIGHT;
+
+// makes api call and returns data
+function getData(dataLink: string) {
+    const apiCall = new XMLHttpRequest();
+    apiCall.open("GET", dataLink, false);
```
JacoBriers: Good idea to change this to use async AJAX call to get some exercise at working with callbacks (since you'll many of them in your time with JavaScript :)

app.ts
```diff
@@ -0,0 +1,147 @@
...
+        // if (firstRecord >= RECORD_COUNT || firstRecord < 0) {
+        alert('please select a NUMBER between 0 and ' + (RECORD_COUNT - 1));
+        return;
+    }
+
+    const contentsDiv = this.content;
+    contentsDiv.innerHTML = ''; // empty table div to fill in a new table
+    const tableElement = new tableGenerator(contentsDiv, firstRecord); // create table using the tableGenerator class
+    tableElement.generateTable();
+    LAST_ENTRY = tableElement.endRecord;
```
JacoBriers: Feels like we're duplicating state here, reckon if you move the `createPage` and `nextPage` methods into the `tableGenerator` class, you could get away with only using the `firstRecord` and `endRecord` members.

app.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const controlsDiv = this.controls;
+    CONTROLS_HEIGHT = controlsDiv.offsetHeight;
+
+    // html for buttons and input text fields
+    controlsDiv.innerHTML += `
+    <button onclick = "createPage(0)">First Page</button>
+    <button onclick = "prevPage()" > Prev </button>
+    <button onclick = "nextPage()" > Next </button>
+    <button onclick = "createPage(document.getElementById(&quot;recordNum&quot;).value)"> Go To </button>
```
JacoBriers: Storing program state in the UI (DOM) is a no-no. Let's find a way around it.

JacoBriers (changes requested):

app.js
nkosimarz: apologies, realized this after commit but didn't know how to remove already committed files,  resolved

app.ts
JacoBriers: Actually I see you're not really storing it in DOM, but prettier to have all your code in JS, not in HTML. You have all the advantages of code-completion and type checking in the JS, none in the HTML.

nkosimarz:
@Winston-Guess @JacoBriers 

refactored to use asynch calls and spinner

app.ts
```diff
@@ -0,0 +1,205 @@
+﻿// Onboarding Project - 2018
+
+let RECORD_COUNT: number; // total number of records available in webserver
+let resizeTimer; // delay required to reduce the number of calls made by browser when resizing the window
+let CONTROLS_HEIGHT;
```
Winston-Guess: Why caps sometimes and other times not?

app.ts
```diff
@@ -0,0 +1,205 @@
...
+let columnsString; // variable to store columns text from api call
+let recordsText; // variable to store records from api call
+let columnsArray;
+
+// makes api call and returns data
+function getRecordCount() {
+	const apiCall = new XMLHttpRequest();
+
+	apiCall.onreadystatechange = function () {
+		if (apiCall.readyState === 4 && apiCall.status === 200) {
```
Winston-Guess: if (!...) return;

app.ts
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+	apiCall.open("GET", "http://localhost:2050/recordCount", true);
+	apiCall.send();
+}
+
+function getCols() {
+	const apiCall = new XMLHttpRequest();
+	apiCall.onreadystatechange = function () {
+		if (apiCall.readyState === 4 && apiCall.status === 200) {
```
Winston-Guess: same as above, and for others below

app.ts
```diff
@@ -0,0 +1,205 @@
...
+
+// makes api call and returns data
+function getRecordCount() {
+	const apiCall = new XMLHttpRequest();
+
+	apiCall.onreadystatechange = function () {
+		if (apiCall.readyState === 4 && apiCall.status === 200) {
+			RECORD_COUNT = JSON.parse(apiCall.response);
+
+			if (columnsString !== undefined) {
```
Winston-Guess: Maybe put a comment here to explain what you're doing.

app.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// hint for the user
+function helpFunc() {
+	alert(
+		"Table for viewing records with IDs 0 to" + (RECORD_COUNT - 1) + " (Numbers Only)."
+		+ "\nTo start from a specific record enter a record ID and click 'Go To'.");
+}
+
+let customPage = () => {
```
Winston-Guess: const

app.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function getTable(start: number, end: number, tableElement) {
+	const apiCall = new XMLHttpRequest();
+
+	const loadDiv = document.createElement('div');
+	loadDiv.setAttribute('class', 'loader')
+	tableElement.myTable.parentNode.appendChild(loadDiv);
+
+	apiCall.onreadystatechange = function () {
```
Winston-Guess: 1. rather use () => { }
2. even better, make this a separate function with a name that describes what it is does.
3. can you not define this function and the parent `getTable` as methods instead?

app.ts
```diff
@@ -0,0 +1,205 @@
...
+		alert('Please type in a number')
+		return;
+	}
+	tableElement.firstRecord = inputVal;
+	tableElement.generateTable();
+}
+
+window.onresize = () => {
+	const doneResizing = () => {
+		tableElement.generateTable();
```
Winston-Guess: Do you need to create a separate function for this?

app.ts
```diff
@@ -0,0 +1,205 @@
...
+
+	clearTimeout(resizeTimer);
+	resizeTimer = setTimeout(doneResizing, 500);
+}
+
+let contentsDiv2: HTMLElement;
+let tableElement: tableGenerator;
+
+// load page for the first time and create buttons
+window.onload = () => {
```
Winston-Guess: I would name this function something like `generateTable` and then go `window.onload = generateTable' then you know what the code is going to do without having to read it first.

app.ts
```diff
@@ -0,0 +1,205 @@
...
+	controlsDiv.innerHTML += `
+ 		<button onclick = "tableElement.firstPage()">First Page</button>
+		<button onclick = "tableElement.prevPage()" > Prev </button>
+		<button onclick = "tableElement.nextPage()" > Next </button>
+		<button onclick = "customPage()"> Go To </button>
+		<input value = "record..." id = "recordNum" >
+		<button onclick = "helpFunc()" > Help </button>`;
+
+	// initialise the table's parent div, create a tableElement
+	contentsDiv2 = document.getElementById('content');
```
Winston-Guess: `...2`?

app.ts
```diff
@@ -0,0 +1,205 @@
...
+		// getTable(this.firstRecord, this.endRecord, this);
+		this.generateTable();
+	}
+
+	// function used by the next page buttton, requires the first and last entry from previously generated table
+	prevPage() {
+		// if the new first entry is less than zero, take user to the first page 
+		this.firstRecord = this.firstRecord - (this.endRecord - this.firstRecord) - 1;
+		this.firstRecord = this.firstRecord < 0 ? 0 : this.firstRecord;
+		// getTable(this.firstRecord, this.endRecord, this);
```
Winston-Guess: left over comments here and above

Winston-Guess (changes requested):
I would make more of the functions methods within a `Table` class

app.ts
nkosimarz: timeout handler will to accept tableElement.generateTable(); unless its wrapped in a function

app.ts
nkosimarz: typo

app.ts
nkosimarz: suggestions don't seem to work

JacoBriers: None of suggestions worked?

app.ts
```diff
@@ -0,0 +1,205 @@
...
+	}
+	apiCall.open("GET", "http://localhost:2050/recordCount", true);
+	apiCall.send();
+}
+
+function getCols() {
+	const apiCall = new XMLHttpRequest();
+	apiCall.onreadystatechange = function () {
+		if (apiCall.readyState === 4 && apiCall.status === 200) {
+			columnsString = apiCall.response;
```
JacoBriers: Looks like this only gets used as a flag to see if AJAX call was successful. In this case, you can make it a boolean flag `isDataFetched`. I would actually use boolean flags throughout for checking if the table is ready to load, nice and consistent.

app.ts
```diff
@@ -0,0 +1,205 @@
...
+
+	constructor(element: HTMLElement, firstRecord: number = 0) {
+		this.element = element;
+		this.firstRecord = firstRecord;
+		this.myTable = document.createElement('table');
+		this.element.appendChild(this.myTable);
+	}
+
+	generateTable() {
+
```
JacoBriers, line 112: I would move the checks if data is loaded in here, and just `return` if the table is not ready to be generated. Function would then be a no-op if required data is not loaded yet.

app.ts
```diff
@@ -0,0 +1,205 @@
+﻿// Onboarding Project - 2018
+
+let RECORD_COUNT: number; // total number of records available in webserver
```
JacoBriers: I would make all these values members of the `TableGenerator` class

app.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// converts strings into array and table
+function parseText(textString: string) {
+	textString = textString.replace(/'/g, '""');
+	const outputArray = JSON.parse(textString);
+	return outputArray;
+}
+
+class tableGenerator {
```
JacoBriers: classes start with uppercase

app.ts
```diff
@@ -0,0 +1,205 @@
...
+function helpFunc() {
+	alert(
+		"Table for viewing records with IDs 0 to" + (RECORD_COUNT - 1) + " (Numbers Only)."
+		+ "\nTo start from a specific record enter a record ID and click 'Go To'.");
+}
+
+let customPage = () => {
+	let inputVal = parseInt(this.recordNum.value);
+	if (isNaN(inputVal)) {
+		alert('Please type in a number')
```
JacoBriers, line 187: Nice

app.ts
```diff
@@ -0,0 +1,205 @@
...
+	CONTROLS_HEIGHT = controlsDiv.offsetHeight;
+
+	// html for buttons and input text fields
+	controlsDiv.innerHTML += `
+ 		<button onclick = "tableElement.firstPage()">First Page</button>
+		<button onclick = "tableElement.prevPage()" > Prev </button>
+		<button onclick = "tableElement.nextPage()" > Next </button>
+		<button onclick = "customPage()"> Go To </button>
+		<input value = "record..." id = "recordNum" >
+		<button onclick = "helpFunc()" > Help </button>`;
```
JacoBriers, line 216: Nice touch. Also like all the options you give the user.

app.ts
```diff
@@ -0,0 +1,205 @@
...
+window.onresize = () => {
+	const doneResizing = () => {
+		tableElement.generateTable();
+	}
+
+	clearTimeout(resizeTimer);
+	resizeTimer = setTimeout(doneResizing, 500);
+}
+
+let contentsDiv2: HTMLElement;
```
JacoBriers: contentsDiv2 does not get used anywhere outside the window.onload function, no need to keep a reference to it

app.ts
```diff
@@ -0,0 +1,205 @@
...
+let tableElement: tableGenerator;
```
JacoBriers: Good idea to put all global state at top of file

app.ts
```diff
@@ -0,0 +1,205 @@
...
+	}
+	apiCall.open("GET", "http://localhost:2050/columns", true);
+	apiCall.send();
+}
+
+function getTable(start: number, end: number, tableElement) {
+	const apiCall = new XMLHttpRequest();
+
+	const loadDiv = document.createElement('div');
+	loadDiv.setAttribute('class', 'loader')
```
JacoBriers: Awesome that this is pure CSS

app.ts
nkosimarz: thank you

PR #29, opened 2018-12-20
Onboarding Project: Please Review

d-munday:


app.ts
```diff
@@ -0,0 +1,153 @@
+﻿let firstIndex:number, lastIndex:number; // stores the indices of the first and last rows in the table
+let recordCount:number; // stores the number of records in the database
+let resizeTimer; // stores time since last window resize
+
+// builds the grid table from the column headers and row data
+function buildTable(columns, records) {
+
+	// create table
+	let table = document.createElement("table");
```
JacoBriers: Consistently use tabs/spaces for formatting. We set up VSCode for our projects to auto-format our code, so this isn't usually a problem, but if you don't have help from the IDE you have to do it manually. Formatting-war is a real thing and we do all we can to prevent it :)

app.ts
```diff
@@ -0,0 +1,153 @@
...
+    for(const header of columns) {
+		var th = document.createElement("th");
+		th.appendChild(document.createTextNode(header));
+		headRow.appendChild(th);
+    };
+    thead.appendChild(headRow);
+	table.appendChild(thead);
+
+	// create rows
+    records.forEach(function (el) {
```
JacoBriers: Rather use `for of` loop or old fashioned for loop

app.ts
```diff
@@ -0,0 +1,153 @@
...
+		recordCount = JSON.parse(data);
+	});
+	
+}
+
+// updates the table display when the window is resized
+// debounce function to reduce frequency of queries made
+window.onresize = () => {
+	clearTimeout(resizeTimer);
+	resizeTimer = setTimeout(function(){
```
JacoBriers: Nice

app.ts
```diff
@@ -0,0 +1,153 @@
...
+function search(e){
+	let input = $("#search").val().toString();
+	let $table = $("tbody tr").toArray();
+
+	for(const record of $table){
+		let td = record.getElementsByTagName("td")[0];
+		record.style.display="block";
+		if(td){
+			let val = td.innerText || td.textContent;
+			if(val.indexOf(input) > -1){
```
JacoBriers: Don't overuse ternary statements but here you could:
```
record.style.display = val.indexOf(input) === -1 ? "none" : "";
```

app.ts
```diff
@@ -0,0 +1,153 @@
...
+		firstIndex = recordCount - calculateRows();
+	} 
+
+	// create loader
+	let loader = document.createElement("loader");
+	document.getElementById("content").appendChild(loader);
+	loader.innerHTML=`<div class="loader" id="loader"></div>`;
+	
+	// outter function to fetch column headers from server
+	$.get("http://localhost:2050/columns", function(columns){
```
JacoBriers: This call will always return the same values, so you can call it once and cache the results, like you did with record count.

app.css
```diff
@@ -0,0 +1,77 @@
...
+.loader{
+	position: absolute;
+	left: 46%;
+	top: 40%;
+	border: 16px solid #f3f3f3;
+	border-radius: 50%;
+	border-top: 16px solid #3498db;
+	width: 120px;
+	height: 120px; 
+	-webkit-animation: spin 2s linear infinite; /* Safari */
```
JacoBriers: Did you test this on Safari?

app.ts
```diff
@@ -0,0 +1,153 @@
...
+				record.style.display = "none";
+			} 		
+		}
+	}
+
+	// if record doesn't exist on current page, query database and display results
+	if(e.which == 13)
+	{
+		let end = parseInt(input) + calculateRows();
+		clearTable();
```
JacoBriers: Since you always `clearTable` before you `loadPage` you can move `clearPage` inside `loadPage`

app.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// clears the table and footer elements
+function clearTable(){
+	document.getElementById("content").innerHTML="";
+	document.getElementById("footer").innerHTML="";
+}
+
+// loads the columns and rows to be displayed based on start and end row indices
+function loadPage(start:number, end:number){
```
JacoBriers: We could make this take only the start index and then you add `caclulateRows()` inside this method to calc the end index.

app.ts
```diff
@@ -0,0 +1,153 @@
...
+	<input type="text" id="search" onkeyup="search(event)" placeholder="Enter ID...">`;
+
+	document.getElementById("loader").remove();
+
+    return table;
+}
+
+// filters rows by user input ID
+function search(e){
+	let input = $("#search").val().toString();
```
JacoBriers, line 49: It isn't a problem here, but when working in the codebase for our other projects, you will need to use more specific div IDs since they are in a "global namespace" of sorts and could conflict with the naming of existing divs.

JacoBriers (changes requested):
Some general comments:
- I like the look and feel, nice and clean
- On the navigation: splitting the search and navigate-to-page functionality that you provide with the "Enter ID" input into two inputs, may make it a bit clearer to the user what they are for.
- Code neatness needs a little work: consistent use of tabs and spaces, use of white-space between variables and operators, etc. There are plugins for IDEs that will help you out here.
Well done!

d-munday:
Thanks for the feedback, Jaco - will submit another pull request in due course once every recommendation has been accounted for.

app.css
d-munday: No I didn't - will remove it for now since it's not necessary to include.

PR #30, opened 2019-06-05
JKG onboarding

jasonkofo:
- Onboarding task
- Was completed without access to the style guide, please forgive any irregularities


app.ts
```diff
@@ -0,0 +1,149 @@
+//#region AppLogic
+//#region API methods
+function getRecordCount() {
+    return fetch('http://localhost:2050/recordCount')
+        .then(function (response) {
```
FritzOnFire: *This should be an arrow function: `(responce) => {`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+            return response.json();
+        });
+}
+function getColumnNames() {
+    return fetch('http://localhost:2050/columns')
+        .then(function (response) {
```
FritzOnFire: *This should be an arrow function: `(responce) => {`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+            return response.json();
+        });
+}
+async function getRecords(fromID: number, toID: number) {
+    const request = 'http://localhost:2050/records?from=' + (fromID - 1).toString() + '&to=' + (toID - 1).toString();
```
FritzOnFire: You dont seem to have any checks for if these values are to big or to small. I seems to try and get -1,36 when I open it the first time and the back-end does not like that :/

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    return await fetch(request)
+        .then(function (response) {
```
FritzOnFire: *This should be an arrow function: `(response) => {`

reinhartfimqs (changes requested):
I am not going to comment on any of the code since you unfortunately did not follow the main instruction that this must be developed in Typescript. The reason for that is that the entire IMQS front end is written in Typescript and the main purpose of this onboarding project is to give you a little exposure to Typescript before jumping into the IMQS www repo.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+function UpdateColumnNames(value: string) {
+    const writable: string = '<th align="center">' + value.toString() + '</th>';
+    $("#HeaderRow").append(writable);
+}
+function placeRecords(fromID: number, toID: number) {
+    $("#ContentBody").empty();
+    getRecords(fromID, toID)
+        .then((records: Array<Array<string>>) => {
+            for (var rowIndex in records) {
```
FritzOnFire: *You should not use `var`, in this case you should use `const`

FritzOnFire: Based on what you are doing with this for-loop it should rather be a `for-of` loop, then you don't have to have that extra index.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                $("#ContentBody").append('<tr id="' + records[rowIndex][0] + '">');
+                for (var columnIndex in records[rowIndex]) {
+                    InsertValue(records[rowIndex][columnIndex], records[rowIndex][0]);
```
reinhartfimqs: Why use a separate when it is not used anywhere else and it is basically a single line.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+function UpdateColumnNames(value: string) {
+    const writable: string = '<th align="center">' + value.toString() + '</th>';
+    $("#HeaderRow").append(writable);
+}
+function placeRecords(fromID: number, toID: number) {
+    $("#ContentBody").empty();
+    getRecords(fromID, toID)
+        .then((records: Array<Array<string>>) => {
+            for (var rowIndex in records) {
+                $("#ContentBody").append('<tr id="' + records[rowIndex][0] + '">');
```
reinhartfimqs: Using jquery to manipulate the DOM is actually an expensive exercise in CPU cycles so I suggest that you do either on append for every row or do one append for all of the row elements.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                for (var columnIndex in records[rowIndex]) {
```
FritzOnFire: *You should not use `var`, in this case you should use `const`

FritzOnFire: Based on what you are doing with this for-loop it should rather be a for-of loop, then you don't have to have that extra index.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    else if (typeof input === "number") {
+        return input;
+    }
+    else {
+        return 0;
+    }
+}
+function nextPageResize(previousCursor: Array<number>): number {
+        $("form-content").height()
```
reinhartfimqs: Is this some left over code, not being used

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        var fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
```
reinhartfimqs: As discussed in person, we basically never use `var` due to the scope it exists in. So, in general, if the value is not going to change or it is a json object we use const. If it is a string or number then we will use let.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    return await fetch(request)
+        .then(function (response) {
+            return response.json();
+        });
+}
+//#endregion
+
+//#region Data Loading methods
+function InsertValue(value: string, id: string) {
+    const writable: string = '<td id=\"' + id + '\" align="center">' + value.toString() + '</td>';
```
FritzOnFire: Not sure what the purpose is of this `id`?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+function nextPageResize(previousCursor: Array<number>): number {
+        $("form-content").height()
+        var fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
+        var toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
+
+        for (var i = fromID; i <= toID; i++) {
+            var element = ($("#" + i.toString()).offset() as JQueryCoordinates).top as number;
+            var documentHeight = ($(window).innerHeight() as number) - ($("form-content").innerHeight() as number) - 10;
```
reinhartfimqs: Why are you subtracting 10? If you put a comment in detailing why you are subtracting a static value and as well as why you use a value of 10 then it would maybe be OK.

But, since you are calculating the height of the document here you should be able to handle that in a css class, look at using calc(100% - 10px) for example in CSS the height.

reinhartfimqs: I see the `$("form-content")` does not have `#`, will this still work without it?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+
+function previousPageResize(previousCursor: Array<number>): Array<number> {
+    var fromId: number = previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]);
+    var toId: number = fromId + step;
+    return [previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]), toId]
+}
+//#endregion
+//#endregion
+
+var step: number = 37;
```
reinhartfimqs: This should be a constant, but more importantly why are you creating a general static variable which is not being used?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    const cursor = [fromID as number, toID as number];
+    $("#HeaderRow").empty();
+    getColumnNames()
+        .then((columns: Array<string>) => {
+            for (var columnIndex in columns) {
+                UpdateColumnNames(columns[columnIndex]);
+            }
+        });
+    placeRecords(fromID, toID);
+    return cursor;
```
FritzOnFire: I think you should just return the values directly, no need to store them in a variable first.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    return [previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]), toId]
+}
+//#endregion
+//#endregion
+
+var step: number = 37;
+var previousCursor: Array<number>;
+
+window.onload = () => {     
+    previousCursor = getPageContent(0,37)
```
reinhartfimqs: I understand why you are using 37 even without a comment which is for a FHD display. What happens when you refresh the page on a 1024x768 resolution display. Essentially you want to calculate the amount of record to display based on the display size

app.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+//#endregion
+
+//#region Handlers
+function getPageContent(fromID: number, toID: number) {
+    const cursor = [fromID as number, toID as number];
+    $("#HeaderRow").empty();
+    getColumnNames()
+        .then((columns: Array<string>) => {
+            for (var columnIndex in columns) {
```
FritzOnFire: *You should not use `var`, in this case you should use `const`

FritzOnFire: You should change this to a `for-of` loop

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                previousCursor = placeRecords(fromId, toId);
+            });
+    });
+}
+var resizeTimer: number = 50;
+$(window).on('resize', function(e) {
+
+    clearTimeout(resizeTimer);
+    resizeTimer = setTimeout(() => {
+        var num = ( $(window).innerHeight() as number )/( $("#10").height() as number );
```
reinhartfimqs: Which element in index.html has a id `10`?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+            if ( element >= documentHeight ){
+                return i;
+            } else {
+                continue
+            };
+        }
+        return toID
+}
+
+function previousPageResize(previousCursor: Array<number>): Array<number> {
```
reinhartfimqs: What happens when you are the first page showing records 1 - 37 and you use the previous page? Do you automatically revert to the last page? Does the reverse apply when you use the next button on the last page?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        });
+    placeRecords(fromID, toID);
+    return cursor;
+}
+
+function toNumber(input: string | number) {
+    if (typeof input === "string") {
+        return parseInt(input);
+    }
+    else if (typeof input === "number") {
```
FritzOnFire: *We prefer to have "cuddled" braces for our `else`s. ei, same line as the closing brace of the `if`

app.ts
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        return input;
+    }
+    else {
+        return 0;
+    }
+}
+function nextPageResize(previousCursor: Array<number>): number {
+        $("form-content").height()
+        var fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
+        var toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
```
FritzOnFire: *This should be a `const`

app.ts
reinhartfimqs: Where is `num` used?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+
+        for (var i = fromID; i <= toID; i++) {
+            var element = ($("#" + i.toString()).offset() as JQueryCoordinates).top as number;
```
FritzOnFire: *This should be a `const`

app.ts
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+            var documentHeight = ($(window).innerHeight() as number) - ($("form-content").innerHeight() as number) - 10;
+
+            if ( element >= documentHeight ){
+                return i;
+            } else {
+                continue
```
FritzOnFire: Missing `;` here.

app.ts
```diff
@@ -0,0 +1,149 @@
...
+            };
```
FritzOnFire: Random `;` here

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        }
+        return toID
```

app.ts
```diff
@@ -0,0 +1,149 @@
...
+//#endregion
+
+//#region Handlers
+function getPageContent(fromID: number, toID: number) {
+    const cursor = [fromID as number, toID as number];
+    $("#HeaderRow").empty();
+    getColumnNames()
+        .then((columns: Array<string>) => {
+            for (var columnIndex in columns) {
+                UpdateColumnNames(columns[columnIndex]);
```
reinhartfimqs: Just consider the amount of DOM calls when appending this way

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                return i;
+            } else {
+                continue
+            };
+        }
+        return toID
+}
+
+function previousPageResize(previousCursor: Array<number>): Array<number> {
+    var fromId: number = previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]);
```
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    var toId: number = fromId + step;
```
FritzOnFire: *This should be a `const`

app.ts
FritzOnFire: *This should be a `const` or a `let`, not sure if you ever change this value...

app.ts
```diff
@@ -0,0 +1,149 @@
...
+    return [previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]), toId]
+}
+//#endregion
+//#endregion
+
+var step: number = 37;
+var previousCursor: Array<number>;
```
FritzOnFire: This should be a `let`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+
+window.onload = () => {     
+    previousCursor = getPageContent(0,37)
+    
+    $("#previous-page").click(() => { 
+        previousCursor = previousPageResize(previousCursor);
+        var fromId: number = (previousCursor[0] >= 0 ? previousCursor[0] : 0) as number;
```
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        var toId: number = (previousCursor[0] >= 0 ? previousCursor[1] : 37) as number;
```
FritzOnFire: *This should be a `const`

app.ts
FritzOnFire: I think you should rather use the `step` value instead of the `37`. You also have a missing `;` here

app.ts
FritzOnFire: I think you should rather use the `step` value instead of the `37`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        previousCursor = placeRecords(fromId, toId);
+    });
+
+    $("#next-page").click(() => {
+        var fromId: number = nextPageResize(previousCursor) as number;
```
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        getRecordCount()
+            .then((value) => {
+                if ( fromId > value - 1 ) {
+                    alert('Step size is too large')
+                } else {
+                    var toId: number = fromId + step < value ? fromId + step : value;
```
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                    previousCursor = placeRecords(fromId, toId);
+                }
+            });
+    });
+
+    $("#go-to-button").click(() => {
+        var index: number = toNumber($("#go-to-index").val() as string);
+        var fromId: number = index;
```
reinhartfimqs: why instantiate `fromId` and then continue to use `index` nearly everywhere else?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        getRecordCount()
+            .then((value) => {
+                var toId: number = (index as number) + step < value ? (index as number) + step : value;
```
reinhartfimqs: Why are you casting index to a number when it already is declared as a number?

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        previousCursor = previousPageResize(previousCursor);
+        var fromId: number = (previousCursor[0] >= 0 ? previousCursor[0] : 0) as number;
+        var toId: number = (previousCursor[0] >= 0 ? previousCursor[1] : 37) as number;
+        previousCursor = placeRecords(fromId, toId);
+    });
+
+    $("#next-page").click(() => {
+        var fromId: number = nextPageResize(previousCursor) as number;
+        getRecordCount()
+            .then((value) => {
```
FritzOnFire: I think this variable needs a beter name than `value`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                if ( fromId > value - 1 ) {
+                    alert('Step size is too large')
```

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                } else {
+                    var toId: number = fromId + step < value ? fromId + step : value;
+                    previousCursor = placeRecords(fromId, toId);
+                }
+            });
+    });
+
+    $("#go-to-button").click(() => {
+        var index: number = toNumber($("#go-to-index").val() as string);
```
FritzOnFire: *This should be a `const`

app.ts
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+        var fromId: number = index;
+        getRecordCount()
+            .then((value) => {
```
FritzOnFire: I think this variable need a beter name than `value`

app.ts
FritzOnFire: *This should be a `const`

app.ts
```diff
@@ -0,0 +1,149 @@
...
+                var toId: number = (index as number) + step < value ? (index as number) + step : value;
+                previousCursor = placeRecords(fromId, toId);
+            });
+    });
+}
+var resizeTimer: number = 50;
```
FritzOnFire: *This should be a `let`

app.ts
FritzOnFire: It doesn't look like this value is being used any more. You should remove it

app.ts
reinhartfimqs: I would think using `fromId` in place of index here would make it more clear what is happening and maybe use a more descriptive name for `value`

FritzOnFire (changes requested):
Comments starting with a `*` I feel are related to the style guide that you did not have access to, so applying the changes is up to you.

As I've mentioned in one of these comments the grid does not work when you open it up anymore as it try to fetch a negative value from the back-end. I have also tested too large values and they fail as well. You should probably add some more sanity checks.

reinhartfimqs (changes requested):
I apologize the non typescript comment earlier, got confused with all of the `var`s being used. There are a couple of things you can look at and then test when you are done next round

reinhartfimqs:
The style guide can be found at `T:\Temp\reinhart\JavaScript Style Guide - Development - Confluence.pdf`

app.ts
```diff
@@ -0,0 +1,171 @@
+//#region AppLogic
+//#region API methods
+function getRecordCount() {
```
JacoBriers: Let's add type definitions for all functions that return values

app.ts
```diff
@@ -0,0 +1,171 @@
...
+        });
+}
+function getColumnNames() {
+    return fetch('http://localhost:2050/columns')
+        .then((response) => {
+            return response.json();
+        });
+}
+async function getRecords(fromID: number, toID: number) {
+    const request = 'http://localhost:2050/records?from=' + (fromID).toString() + '&to=' + (toID).toString();
```
JacoBriers: You can use something called a template string to compose these type of string. See this link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

app.ts
```diff
@@ -0,0 +1,171 @@
...
+            return response.json();
+        });
+}
+//#endregion
+
+//#region Data Loading methods
+
+function placeRecords(fromID: number, toID: number) {
+    getRecords(fromID, toID)
+        .then((records: Array<Array<string>>) => {
```
JacoBriers: We typically use the `string[][]` syntax instead of the `Array<Array<string>>`.

app.ts
```diff
@@ -0,0 +1,171 @@
...
+                appendable += '</tr>';
+                
+            }
+            $("#ContentBody").empty();
+            $("#ContentBody").append(appendable);
+        });
+    return [fromID as number, toID as number];
+}
+
+// Credit: https://gist.github.com/ca0v/73a31f57b397606c9813472f7493a940
```
JacoBriers: I like your honesty

app.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+    else {
+        return 0;
+    }
+}
+function nextPageResize(previousCursor: Array<number>): number {
+        let fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
+        let toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
+
+        const documentHeight = ($(window).innerHeight() as number) - ($("#" + fromID.toString()).height() as number);
```
JacoBriers: Although HTML5 does allow you to start a div `id` with a number, it's good practice to make you id as descriptive as possible. Div ids need to be unique across the entire DOM (otherwise weird things start happening). This means it's a good idea prefixing them with some descriptive string like `table-row-42`.

app.ts
```diff
@@ -0,0 +1,171 @@
...
+                        let toId: number = (fromId as number) + step < value ? (fromId as number) + step : value - 1;
+                        previousCursor = placeRecords(fromId, toId);
+                    }
+                } else {
+                    alert('It seems you are not inserting an integer - please ensure that you are.');
+                }
+            });
+    });
+}
+let resizeTimer: number = 50;
```
JacoBriers: I would assign this to -1 or 0 to make it seem a bit more like an "empty" variable.

app.ts
JacoBriers: Agreed. 

On my screen it looks like this for instance:
image

If I remove the height CSS on the body:
image

package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "1.0.0",
+  "description": "Onboarding project",
+  "main": "index.js",
+  "dependencies": {},
+  "devDependencies": {
+    "@types/jquery": "^3.3.29"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "tsc -w --project ./"
```
JacoBriers, line 13: I'm guessing TypeScript was installed as a global node package on your PC. If you want to install it locally (to this project), run `npm install --save-dev typescript`. This will add typescript to the dependencies of this project, making the version of typescript explicit for any other devs that may work on it. Like we spoke earlier, this is probably not something that is that important to a small on-boarding project like this, but on anything bigger it is good to lock this down.

JacoBriers (changes requested):

index.html
```diff
@@ -3,11 +3,64 @@
...
+	display: none;
+	}
+	tr {
+	 border: 1px solid #ccc; 
+	}
+	td {
+	border: none;
+	border-bottom: 1px solid #eee;
+	position: relative;
+	padding-left: 50%;
```
Winston-Guess: Formatting

index.html
```diff
@@ -3,11 +3,64 @@
...
-<body>
-	<p>Hello</p>
+<body id="body" style="overflow: hidden; height: 5002px">
+	<div id="form-content">
+		Go To Start Index: <input type="number" id="go-to-index" value="0">
+		<button id="go-to-button" type="button">Go to Index</button>
+		<button id="previous-page" type="button">Go to Previous Page</button>
+		<button id="next-page" type="button">Go to Next Page</button>
+	</div>
+	<table id='WrapperTable' style="width:100%; height:100%; table-layout: fixed;">
```
Winston-Guess: Try keep things consistent with formatting. Our convention is to use dashes for `id` and `class` i.e. `id='wrapper-table'`

app.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+    else if (typeof input === "number") {
+        return input;
+    }
+    else {
+        return 0;
+    }
+}
+function nextPageResize(previousCursor: Array<number>): number {
+        let fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
```
Winston-Guess: Most of your `let` variables should be `const`. `let` signifies that you are making a variable that could be adjusted later down the line. In the case of an array or an object a `const` signifies that the variable is not going to be reassigned to a completely new object or array.

app.ts
```diff
@@ -0,0 +1,171 @@
...
+        let toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
+
+        const documentHeight = ($(window).innerHeight() as number) - ($("#" + fromID.toString()).height() as number);
+        for (let i = fromID; i <= toID; i++) {
+            let element = ($("#" + i.toString()).offset() as JQueryCoordinates).top as number;
```
Winston-Guess: `elementHeight` would be a more apt naming here

app.ts
Winston-Guess: You could actually remove all the `else` and `else if` usage since the `return` takes care of that. This is stylistic though, not sure what @FritzOnFire and @JacoBriers think?

app.ts
```diff
@@ -0,0 +1,171 @@
+//#region AppLogic
+//#region API methods
+function getRecordCount() {
+    return fetch('http://localhost:2050/recordCount')
+        .then((response) => {
+            return response.json();
+        });
+}
+function getColumnNames() {
```
Winston-Guess: This is the more modern approach:
```
async function getUserAsync(name) {
  const response = await fetch(`https://api.github.com/users/${name}`);
  return await response.json();
}
```

app.ts
```diff
@@ -0,0 +1,171 @@
...
+        getRecordCount()
+            .then((value) => {
+                fromId = fromId == value - 1 ? fromId - step : fromId;
+                toId = toId <= value - 1 ? toId : value - 1;
+                previousCursor = placeRecords(fromId, toId);
+            });
+        
+    });
+
+    $("#next-page").click(() => {
```
Winston-Guess: You can do the `async` `await` thing here as well. Let me know if you need a quick intro into it. There is a lot to read on the subject but I find that it's majorly over-complicated..

Winston-Guess (changes requested):

app.ts
JacoBriers: Hmmm... my preference here would be a case statement. But yes, because we return, we don't need the else.

app.ts
JacoBriers: Agreed, the code feels nice and synchronous if you do this

app.ts
Winston-Guess: 👍 @JacoBriers. I try not to think about switches in JavaScript unless I have more than 3/4 cases because they fallthrough by default 😢 However, I realise this can be circumvented by placing them in their own function and using returns to make things look nicer. In this specific case a switch is perfect because that isn't even necessary 😄

app.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+function previousPageResize(previousCursor: number[]): number[] {
+    const toId = calculateToId(previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]));
+    return [previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]), toId];
+}
+//#endregion
+//#endregion
+
+let previousCursor: number[];
```
Winston-Guess: Variables that do not form part of any class or function should go at the top of the page (below imports)

app.ts
```diff
@@ -0,0 +1,175 @@
+//#region AppLogic
+//#region API methods
+async function getRecordCount() : Promise<number> {
+    const response = await fetch('http://localhost:2050/recordCount');
+    return await response.json();
+}
+async function getColumnNames() : Promise<string[]>{
```
Winston-Guess, line 15: Place an empty line between each class or function

app.ts
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+        return toID;
+}
+
+function previousPageResize(previousCursor: number[]): number[] {
+    const toId = calculateToId(previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]));
+    return [previousCursor[0] - (nextPageResize(previousCursor) - previousCursor[0]), toId];
+}
+//#endregion
```
Winston-Guess: I like these comments but I would include a description of what region is ending

app.ts
```diff
@@ -0,0 +1,175 @@
...
+
+let previousCursor: number[];
+let previousScale: number;
+
+window.onload = async () => {     
+    previousCursor = await getPageContent(0, calculateToId(0));
+    
+    $("#previous-page").click(async () => { 
+        previousCursor = previousPageResize(previousCursor);
+        let fromId = (previousCursor[0] >= 0 ? previousCursor[0] : 0);
```
Winston-Guess: You don't need the brackets here and in some of your other ternaries

```suggestion
        let fromId = previousCursor[0] >= 0 ? previousCursor[0] : 0;
```

index.html
```diff
@@ -3,11 +3,69 @@
...
+
+		</tbody>
+		</table>
+	<div id="table-arbitrary-footer" ></div></div>
 </body>
+<style>
+	/* Credit: https://stackoverflow.com/questions/37723448/how-to-make-responsive-table-without-using-bootstrap/37723905 */
+	table tr td,table tr td{
+		text-align: center;
+	}
```
Winston-Guess, line 30: Empty line here too

app.ts
```diff
@@ -0,0 +1,175 @@
...
+        });
+    return await placeRecords(fromID, toID);
+}
+
+function toNumber(input: string | number, output: any = 1) : number {
+    switch (typeof input) {
+        case ('string'):
+            if (output == 1) {
+                return parseInt(input as string);
+            } else {
```
Winston-Guess: You can still remove the else here so that `return parseFloat(input as string);` is unindented here

app.ts
```diff
@@ -0,0 +1,175 @@
...
+        recordDisplayOffset = 0
+    } else if (window.innerHeight <= 969) {
+        recordDisplayOffset = 1;
+    } else if (window.innerHeight <= 1938) {
+        recordDisplayOffset = 3
+    } else {
+        recordDisplayOffset = 15
+    }
+
+    return recordDisplayOffset + (fromId + Math.floor((window.innerHeight - ($("#form-content").innerHeight() as number)) / 37));
```
Winston-Guess: There is a little too much logic in this one line for me, being succinct is not good idea when it makes readability difficult. You're going 5 brackets deep in one line here. Rather separate things in to one or two more variables with variable names that describes the logic. A few good variable names often make commenting unnecessary.

app.ts
```diff
@@ -0,0 +1,175 @@
...
+    });
+}
+let resizeTimer = 50;
+window.onresize = () => {
+    const x = calculateToId(previousCursor[0]);
+    clearTimeout(resizeTimer);
+    resizeTimer = setTimeout(async () => {
+        const recordCount = await getRecordCount();
+        if ( x >= recordCount - 1 ) {
+            previousCursor = await placeRecords(recordCount - 1 - (calculateToId(previousCursor[0]) - previousCursor[0]), recordCount - 1);
```
Winston-Guess: Try shorten long line here

app.ts
```diff
@@ -0,0 +1,175 @@
...
+        recordDisplayOffset = 15
+    }
+
+    return recordDisplayOffset + (fromId + Math.floor((window.innerHeight - ($("#form-content").innerHeight() as number)) / 37));
+}
+
+function nextPageResize(previousCursor: number[]): number {
+        const fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
+        const toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
+        const documentHeight = ($(window).innerHeight() as number) - ($(`#table-row-${fromID.toString()}`).height() as number);
```
Winston-Guess: remove extraneous brackets and number casting

app.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        for (let i = fromID; i <= toID; i++) {
+            const elementHeightOffset = ($(`#table-row-${i.toString()}`).offset() as JQueryCoordinates).top as number;
```
Winston-Guess: do you need all the casting here?

app.ts
```diff
@@ -0,0 +1,175 @@
...
+                    let toId = (fromId) + possibleStep < recordCount ? (fromId) + possibleStep : recordCount - 1;
+                    previousCursor = await placeRecords(fromId, toId);
+                }
+            } else {
+                alert('It seems you are not inserting an integer - please ensure that you are.');
+            }
+        }
+    });
+}
+let resizeTimer = 50;
```
Winston-Guess: move to top

Winston-Guess (changes requested):
Good work! It is looking a lot cleaner. Just a few more stylistic adjustments could be made.

app.ts
jasonkofo: casting is needed as JQuery's .innerHeight() method could be undefined

Removing the bracketsz causes this error
image

jasonkofo: Specifically for that term and not so much for the first term.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+
+function nextPageResize(previousCursor: number[]): number {
+        const fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
+        const toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
+        const documentHeight = $(window).innerHeight() as number - ($(`#table-row-${fromID.toString()}`).height() as number);
+
+        for (let i = fromID; i <= toID; i++) {
+            const elementHeightOffset = ($(`#table-row-${i.toString()}`).offset() as JQueryCoordinates).top;
+
+            if ( elementHeightOffset >= documentHeight ){
```
Winston-Guess: No need for spaces but formatter would sort that out
```suggestion
            if (elementHeightOffset >= documentHeight){
```

index.html
```diff
@@ -3,11 +3,75 @@
...
+			width: 45%;
+			padding-right: 10px;
+			text-align:left;
+			font-weight: bold;
+		}
 
+		td:before { 
+			content: attr(data-title);
+		}
+}
```
Winston-Guess: Being nit picky but this braces is off. Again formatter would sort this out.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+        case ("number"):
+            return input as number;
+        default:
+            return 0;
+    }
+}
+
+function calculateToId(fromId: number): number {
+    const possibleRecords = Math.floor((window.innerHeight - ($("#form-content").innerHeight() as number)) / 37);
+    const possibleId = (fromId + possibleRecords);
```
Winston-Guess: bracket not necessary

Winston-Guess (approved):
LGTM! Just a few last nit-picky comments but check with @JacoBriers if he is happy for you to leave them out of this PR and move on 😄👍

app.ts
```diff
@@ -0,0 +1,187 @@
...
+    return await response.json();
+}
+
+async function getColumnNames() : Promise<string[]>{
+    const response = await fetch('http://localhost:2050/columns');
+    return await response.json();
+}
+
+async function getRecords(fromID: number, toID: number): Promise<string[][]> {
+    const response = await fetch(`http://localhost:2050/records?from=${(fromID).toString()}&to=${(toID).toString()}`);
```
JacoBriers: No need to use `toString()` when using template strings.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+    const records = await getRecords(fromID, toID)
+    let appendable = '';
+    for (const record of records) {
+            appendable += `<tr id="table-row-${record[0]}">`;
+        for (const column of record) {
+            appendable += `<td align="center">${column.toString()}</td>`;     
+        }
+        appendable += '</tr>';
+    }
+    $("#WrapperTable-ContentBody").empty();
```
JacoBriers: Use either camelcase or dashes, not both

app.ts
```diff
@@ -0,0 +1,187 @@
...
+    $("#WrapperTable-ContentBody").append(appendable);
+    return await [fromID as number, toID as number];
```
JacoBriers: `fromID` and `toID` do not get mutated in this method, why pass them into the function and then return them as-is?

JacoBriers: Also, no need to `await` here

app.ts
```diff
@@ -0,0 +1,187 @@
...
+        .then((columns: string[]) => {
+            for (const column of columns) {
+                let writable: string = `<th align="center">${column}</th>`;
+                $("#WrapperTable-HeaderRow").append(writable);
+            }
+        });
+    return await placeRecords(fromID, toID);
+}
+
+function toNumber(input: string | number, output: any = 1) : number {
```
JacoBriers: I find the `output` parameter a bit weird. It looks like a flag to either parse the value as an integer or a float. But then a better way to do this would be to name it `parseAsFloat` and declare it as a `boolean`.

The only time I see the flag used for parsing a float was for user-input. I don't see a user typing in a floating point number as an index; would actually expect an error message if that was the case?

app.ts
```diff
@@ -0,0 +1,187 @@
...
+            return parseFloat(input as string);
+        case ("number"):
+            return input as number;
+        default:
+            return 0;
+    }
+}
+
+function calculateToId(fromId: number): number {
+    const possibleRecords = Math.floor((window.innerHeight - ($("#form-content").innerHeight() as number)) / 37);
```
JacoBriers, line 78: Instead of casting, it is better to check for `undefined`. In this case it is safe because the div would always exist, but in our production code we would never assume this. Better to check explicitly like this:
```
const headerHeight = $("#form-content").innerHeight();
if (!headerHeight) {
    alert("Some error message");
    return;
}
const possibleRecords = Math.floor((window.innerHeight - headerHeight) / 37);
```

Casting is basically a way to get around/ignore the TypeScript type system, so only use it as a last resort.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+    const possibleId = fromId + possibleRecords;
+
+    let recordDisplayOffset = 0;
+    if (window.innerHeight <= 646) {
```
JacoBriers, line 82: This seems very arbitrary, there will be a way to not hard-code these offsets or to calculate it using the row height and the total table height.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+
+function nextPageResize(previousCursor: number[]): number {
+        const fromID = toNumber(previousCursor.sort((a, b) => {return a - b})[0]);
+        const toID = toNumber(previousCursor.sort((a, b) => {return a - b})[1]);
+        const documentHeight = $(window).innerHeight() as number - ($(`#table-row-${fromID.toString()}`).height() as number);
+
+        for (let i = fromID; i <= toID; i++) {
+            const elementHeightOffset = ($(`#table-row-${i.toString()}`).offset() as JQueryCoordinates).top;
+
+            if (elementHeightOffset >= documentHeight){
```
JacoBriers: For a bit less nesting you can:
```
if (elementHeightOffset < documentHeight) continue;
return i;
```

JacoBriers (commented):
No need to implement these changes, just go through them and let me know if something is not clear

app.ts
jasonkofo: At some point, Fritz mentioned about inserting sanity checks to make sure that the user does not insert any unwanted values into the field that could break the API call.

I imagined a scenario in which he would've tested to make sure that someone does not insert this breaking change - but I can remove it just as readily.

JacoBriers: Jip, agree you need to show an error message when the user types in something that doesn't make sense. But I would lean towards not having the `output` parameter at all and ignoring the decimal bit and just parsing the number as an int.

JacoBriers (approved):

PR #31, opened 2019-11-14
Onboarding finished

YuveerMaharaj:


app.ts
```diff
@@ -0,0 +1,143 @@
+//Loads the table to the screen
+window.onload = () => GetColumnNames();
+
+//Variable declarations
+let totalRecordCount;
```
jasonkofo: Just as a matter of design, I would rather you not use these methods to write to global variables. It's difficult to see the danger of using global variables in the scope of this onboarding project; but imagine if you had 100 different files or editing the same variable (if you had chosen to export this variable and make it truly global) - or even if the file was bigger and all the different sections of the code were all making changes to this one lonely variable. Imagine the bugs that would break out right through your system.

The issue is not even that you are using them in this space, because here I would suppose they are fine, but the issue is that your methods are making changes to the same variable. While in Typescript/Javascript, this may not necessarily be a problem because the language is single threaded (you will learn what that means as you grow in your career), in other languages this could prove very challenging to solve bugs.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+let records: any[];
+let from: number = 1;
+let to: number = 30;
+
+//Function to retrieve total amount of records
+function GetNumberOfRecords() {
+  $.ajax({
+    url: "/recordCount",
+    type: "GET",
+    timeout: 12000,
```
jasonkofo, line 20: Quite a long timeout you've got there, considering the server runs on the same machine as your front-end code - but I must commend the awareness that you are demonstrating.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+      records = responseRecords;
+      NextSync();
+    })
+    .fail(function() {
+      alert("No available records for your selection");
+    });
+}
+
+//Function to sync API calls
+function Sync() {
```
jasonkofo: I don't understand why you're wrapping this single method in another single method? Why not just call _getActualRecords_ directly?

app.ts
```diff
@@ -0,0 +1,143 @@
...
+  GetActualRecords();
+}
+
+//Function to sync API calls
+function NextSync() {
```
jasonkofo: Same as above, why not just call _BuildTable_ directly?

app.ts
```diff
@@ -0,0 +1,143 @@
...
+
+//Pages to the next set of data in increments of 30
+function NextPage() {
+  from = from + 30;
+  to = to + 30;
+  GetActualRecords();
+}
+
+//Pages backwards through data in increments of 30
+function PreviousPage() {
```
jasonkofo, line 125: once again, these methods make calls to the global variables in this scope even though it isn't class based.

app.ts
```diff
@@ -0,0 +1,143 @@
+//Loads the table to the screen
```
jasonkofo, line 1: When you want install npm packages, and you want other teammates to make use of these packages, you need to run
```
npm install --save <package-name>
```
This saves the package into your package.json, so that when anyone who uses this repo runs 'npm install', they will get your 3rd party libraries that you used to make your source code work.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+
+//Function to retrieve total amount of records
+function GetNumberOfRecords() {
+  $.ajax({
+    url: "/recordCount",
+    type: "GET",
+    timeout: 12000,
+    dataType: "text"
+  })
+    .done(function(responseText: any) {
```
jasonkofo, line 23: Thanks for using type annotations. That's one of the reasons why Typescript (and other Javascript supersets) is so popular and necessary in our days and age. We add these _loosely_ fitting type declarations so that we as developers "write better code". Unlike other strongly typed languages, these type annotations do not mean anything substantive in Typescript, but they add a perceived layer of type safety that when debugging (in my experience), improves the process.
In Javascript, which Typescript eventually gets compiled to, objects are just that - objects. Whatever abstraction of these objects we add does not actually do anything, which is why these are called type _annotations_, rather than _declarations_. But it helps other people read the code better from a multitude of different angles.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+      totalRecordCount = responseText;
+    })
+    .fail(function() {
```
jasonkofo, line 29: I would rather you use the arrow function here. That looks something like 
```ts
    .fail(() => {
```
For your purpose at the moment, this does not do anything really - so I'm not going to explain the significance. 
Just know that at IMQS, we prefer these arrow functions.

This request to change most of these to arrow functions applies to basically every internally declared function, and I will comment on it just now.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+//Removes current table to enable pagination (Called in build table function)
+function RemoveTable() {
+  if (document.getElementById("Table") != null) {
+    $("Table").remove();
+  }
+}
+
+//Activates the submit button to produce users search results
+function SearchBar() {
+  let userInput = <HTMLInputElement>document.getElementById("userInput");
```
jasonkofo, line 143: I believe you can change the _let_ to a _const_ as below
```ts
   const userInput = <HTMLInputElement>document.getElementById("userInput");
```
My reason for saying this is that using _let_ communicates to us that the _userInput_ variable is going to change in the coming scope of the method, which I cannot really say that it will (you only seem to be reading the _value_ property from it).
On the immediate opposite side is the use of the _const_ keyword, which lets us know that you are not interested in changing the value of that variable - which you don't.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+  from = parseInt(userInput.value);
+  console.log("userInput: " + from);
+  to = from + 30;
+
+  GetActualRecords();
+}
+
+//Enables mouse hover over table rows for readability
+$("td").hover(
```
jasonkofo: You have already done this in the css for this project, I happened upon this when I realized that none of these styles were actually being applied to the components,

Please remove this section given that it doesn't seem to do anything.

app.css
```diff
@@ -0,0 +1,97 @@
...
+	background-color: blueviolet;
+	border: none;
+	color: white;
+	padding: 16px 32px;
+	text-align: center;
+	text-decoration: none;
+	display: inline-block;
+	font-size: 16px;
+	margin: 4px 2px;
+	-webkit-transition-duration: 0.4s; /* Safari */
```
jasonkofo, line 101: Good on you for using Styles that work across different browsers :)

app.css
```diff
@@ -0,0 +1,97 @@
+/* table{width: 100%; border: 2px black;} */
```
jasonkofo: Maybe consider removing the styles that are no longer relevant, I'll let you decide which ones they are

app.ts
```diff
@@ -0,0 +1,143 @@
...
+
+//function to retrieve column names
+function GetColumnNames() {
+  $.ajax({
+    url: "/columns",
+    type: "GET",
+    timeout: 12000,
+    dataType: "JSON"
+  })
+    .done(function(responseJSON: any) {
```
jasonkofo: replace with arrow function

app.ts
```diff
@@ -0,0 +1,143 @@
...
+      columnNames = responseJSON;
+      Sync();
+    })
+    .fail(function() {
```
jasonkofo: replace with arrow function

app.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+//Function to retrieve the record data
+function GetActualRecords() {
+  $.ajax({
+    url: "/records?from=" + from + "&to=" + to,
+    timeout: 12000,
+    dataType: "JSON"
+  })
+    .done(function(responseRecords: any) {
```
jasonkofo: replace with arrow function

app.ts
```diff
@@ -0,0 +1,143 @@
...
+      records = responseRecords;
+      NextSync();
+    })
+    .fail(function() {
```
jasonkofo: replace with arrow function

app.ts
```diff
@@ -0,0 +1,143 @@
...
+    });
+}
+
+//function to retrieve column names
+function GetColumnNames() {
+  $.ajax({
+    url: "/columns",
+    type: "GET",
+    timeout: 12000,
+    dataType: "JSON"
```
jasonkofo: Let's use lower case 'json' rather than upper case 'JSON'.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+      alert("Could not display column names");
+    });
+}
+
+//Function to retrieve the record data
+function GetActualRecords() {
+  $.ajax({
+    url: "/records?from=" + from + "&to=" + to,
+    timeout: 12000,
+    dataType: "JSON"
```
jasonkofo, line 56: lower case "json" rather than upper case "JSON"

app.ts
```diff
@@ -0,0 +1,143 @@
...
+  const body = document.getElementsByTagName("body")[0];
+
+  const ourTable = document.createElement("table");
+  ourTable.setAttribute("id", "Table");
+
+  const thead = document.createElement("thead");
+  ourTable.appendChild(thead);
+
+  //Column names are created for the table
+  for (let c = 0; c < columnNames.length; c++) {
```
jasonkofo: Let's make c a const, since at the end of every iterative loop it gets re-assigned.
```ts
   for (const c = 0; c < columnNames.length; c++) {
```
It is a bit trippy at first, but for the sake of convention.

jasonkofo (changes requested):

app.ts
Winston-Guess: arrow function

app.ts
```diff
@@ -0,0 +1,193 @@
...
+    from = from + NumberOfRows;
+    to = to + NumberOfRows;
+    //Resize();
+    GetActualRecordsAmount();
+  }
+}
+
+//Pages backwards through data
+function PreviousPage() {
+  let NumberOfRows = Math.floor((window.innerHeight - 50) / 24) - 1;
```
Winston-Guess, line 126: `const` same applies to quite a few other `let`s

For reference: https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75

app.ts
```diff
@@ -0,0 +1,193 @@
...
+      to = from + NumberOfRows;
+    }
+
+    console.log("userInput: " + from);
+    GetColumnNames();
+  }, 500);
+}
+
+function restrictAlphabets(e: any) {
+  var x = e.which || e.keycode;
```
Winston-Guess, line 189: `var` and `x` isn't a great name

app.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  //Column names are created for the table
+  for (let c = 0; c < columnNames.length; c++) {
+    thead
+      .appendChild(document.createElement("th"))
+      .appendChild(document.createTextNode(columnNames[c]));
+  }
+
+  //Cells are created and populated with data
+  for (let i = 0; i < records.length; i++) {
```
Winston-Guess, line 91: Use new for loop methods from style guide

i.e. `for (const key in records) { ... }`
or `for (const record of records) { ... }`

app.ts
```diff
@@ -0,0 +1,193 @@
+//Loads the table to the screen
+window.onload = () => GetTotalNumberOfRecords();
+//Resizes page during window resizing
+window.onresize = () => Resize();
+
+//Variable declarations
+let totalRecordCount: number;
+let columnNames: any[];
```
Winston-Guess, line 8: `const` see comment further down

Winston-Guess (changes requested):

PR #32, opened 2019-11-18
Helaine-Onboarding

He7i:
Have mercy on my soul


app.ts
```diff
@@ -0,0 +1,130 @@
...
+
+const rowHeight = 56;
+let from = 0;
+let to = 0;
+
+
+function amountOfRowsThatFit(){
+	let $buttonEl = $("button");
+
+	let windowHeigt = window.innerHeight;
```
Winston-Guess: All fields that do not get changed within their scope should be `const`, objects and arrays should be `const`s in most cases too

app.ts
```diff
@@ -0,0 +1,130 @@
...
+	let $from = $("toValue");
+	callHeaders();
+	from = 0;
+	let rowsFittable =amountOfRowsThatFit();
+	to = from+rowsFittable;
+
+}
+
+
+function pressedNext(){
```
Winston-Guess: You need to sort out formatting throughout this PR and you should use a class or classes and methods as opposed to just functions. And when you create functions within methods use arrow functions, i.e. `() => {...}`

app.ts
```diff
@@ -0,0 +1,130 @@
...
+function create(input: string[]) {
+	//create table
+	let tableRetrieved = <HTMLInputElement>document.getElementById("intialTable");
+	console.log(tableRetrieved);
+	if (tableRetrieved != null) {
+		tableRetrieved.remove();
+	}
+
+
+	let table = document.createElement('table');
```
Winston-Guess: As I said above, use `const` here since it is an object and the object is never going to be reassigned to another object, array, or value...

app.ts
```diff
@@ -0,0 +1,130 @@
...
+	fittableRows = retrieveRowsThatCanFit();
+	callHeaders();
+
+
+}
+
+
+function create(input: string[]) {
+	//create table
+	let tableRetrieved = <HTMLInputElement>document.getElementById("intialTable");
```
Winston-Guess, line 152: New and better way of doing casting IMO
```suggestion
	let tableRetrieved = document.getElementById("intialTable") as HTMLInputElement;
```

app.ts
```diff
@@ -0,0 +1,130 @@
...
+	let indexNumber = 0;
+
+	//validate input
+	let rowsRequest = new XMLHttpRequest;
+	//Call the open function, GET-type of request,
+	rowsRequest.open('GET', 'http://localhost:2050/records?from=' + from + '&to=' + to, true)
+
+	let generateRows = () => {
+		//check if the status is 200(means everything is okay)
+		let columnData = Array.from(rowsRequest.responseText.split(','));
```
Winston-Guess: There is something going wrong with your formatting here so I can't quite figure out what's happening here. There is definitely a problem with tabbing and/or bracing

app.ts
```diff
@@ -0,0 +1,130 @@
...
+
+	from = ($from.value)-rowsThatFit;
+     to = from-1;
+ 
+
+}
+
+
+
+//retrieves column names
```
FritzOnFire: Use the java-doc style comments:
```ts
/**
 * Retrieves column names
 */
```

Winston-Guess (changes requested):
I've given you a number of general recommendations that you can use throughout your PR. Once this is ready for a more in depth review let us know :+1:

app.ts
```diff
@@ -0,0 +1,130 @@
+"use strict";
+window.onload = () => {onloadValues() };
+
+const rowHeight = 56;
+let from = 0;
+let to = 0;
+
+
+function amountOfRowsThatFit(){
```
nkosimarz: I think you need to specify the return type here

app.ts
```diff
@@ -0,0 +1,130 @@
...
+//retrieves column names
+function callHeaders() {
+	let xhr = new XMLHttpRequest;
+	//Call the open function, GET-type of request,
+	xhr.open('GET', 'http://localhost:2050/columns', true)
+	//call the onload
+	xhr.onload = function () {
+		//check if the status is 200(means everything is okay)
+		if (this.status === 200) {
+			create(Array.from(JSON.parse(this.responseText)));
```
nkosimarz: You could make this one line

app.ts
```diff
@@ -0,0 +1,130 @@
...
+	fittableRows = retrieveRowsThatCanFit();
+	callHeaders();
+
+}
+
+function pressedPrevious() {
+	fittableRows = retrieveRowsThatCanFit();
+	callHeaders();
+
+
```
nkosimarz, line 91: A few unnecessary lines, can you delete them?

app.ts
```diff
@@ -0,0 +1,130 @@
...
+     to = from-1;
+ 
+
+}
+
+
+
+//retrieves column names
+function callHeaders() {
+	let xhr = new XMLHttpRequest;
```
FritzOnFire: You should rather use `fetch`. Something like this:
```ts
fetch('http://localhost:2050/columns')
    .then(res => {
        if (res.status != 200) {
            // Bleh
            return;
        }
        return res.json(); // This already does JSON.parse
    })
    .then(json => {
        // Do fun json stuff
    })
```

app.ts
```diff
@@ -0,0 +1,130 @@
...
+
+function retrieveRows() {
+	let indexNumber = 0;
+
+	//validate input
+	let rowsRequest = new XMLHttpRequest;
+	//Call the open function, GET-type of request,
+	rowsRequest.open('GET', 'http://localhost:2050/records?from=' + from + '&to=' + to, true)
+
+	let generateRows = () => {
```
FritzOnFire: Where is this functions closing brace?

nkosimarz (changes requested):

app.ts
FritzOnFire: *hard casting*. Also, take note of this comment, but don't apply it. There is much history in what @Winston-Guess is trying to convince you to do, but just stick to the style guide and you will be fine.

app.ts
```diff
@@ -0,0 +1,130 @@
...
+	callHeaders();
+
+
+}
+
+
+function create(input: string[]) {
+	//create table
+	let tableRetrieved = <HTMLInputElement>document.getElementById("intialTable");
+	console.log(tableRetrieved);
```
FritzOnFire: Stray console log

app.ts
```diff
@@ -0,0 +1,130 @@
...
+	if (tableRetrieved != null) {
+		tableRetrieved.remove();
+	}
+
+
+	let table = document.createElement('table');
+	table.setAttribute('id', 'intialTable');
+	document.getElementById("body")!.appendChild(table);
```
FritzOnFire, line 158: I think you should first build up your table fully before you append it to body, all those `getElementById` calls are expensive.

index.html
```diff
@@ -3,11 +3,24 @@
...
-<body>
-	<p>Hello</p>
-</body>
 
-</html>
+<div class="buttons" id="buttons">
+
+	<button type="button" id="previousPage"  onclick="pressedPrevious()">Previous</button>
+	<button type="button" id="nextPage" onclick="pressedNext()">Next</button>
+	<script type="text/javascript" src="./app.js"></script>
```
FritzOnFire: This sould be in your `<head>`

tsconfig.json
```diff
@@ -17,6 +17,6 @@
 		],
 	},
 	"include": [
-		"./**/*.ts"
+		"./**/*.ts",
```
FritzOnFire: Remove the `,`

FritzOnFire (changes requested):
Your formatting seems to be a bit wierd. It looks like your ide is not formatting your code. Im expecting to see spaces before your `{` when you declare a function and spaces around `+`s. There also seems to be a mix of tabs, spaces and tabs with spaces.

Also I can't seem to find the code that actually build's up your table? All I could find was the table headers being built.

app.ts
Winston-Guess: If the conditional functionality is complex I would keep it multiple lines and keep to the braces. However, in this case I feel it would be best to do the following
```suggestion
			if (this.status !== 200) return; // in real life log (and/or notify user of) the error
			create(Array.from(JSON.parse(this.responseText)));
```

app.ts
jasonkofo: Also, pls correct the spelling of _Height_

app.ts
```diff
@@ -0,0 +1,187 @@
+"use strict";
+window.onload = () => { onloadValues() };
```
jasonkofo: Rather write it like this - burying this single method call does not read cleanly.
```ts
window.onload = onloadValues();
```
Also, my personal preference is that you call this _onLoadValues_ (note the casing).

app.ts
```diff
@@ -0,0 +1,187 @@
...
+window.onresize = () => { fromAndToValuesAdjustment() };
```
jasonkofo: Same as above
```ts
window.onresize = fromAndToValuesAdjustment();
```

app.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const rowHeight = 49 + 15;
```
jasonkofo: You've made this a const (which in Typescript disallows your from editing the value later), but you are adding two numbers?

Also - I am assuming from the previous messages that you have access to the style guide. Pls advise me, does it say something about declaring all of your variables at the top of the file/scope?

app.ts
```diff
@@ -0,0 +1,187 @@
...
+let from = 0;
+let to = 0;
+let fittableRows = 0;
+
+/**
+ * calculates the amount of rows that can fit on the screen given its current height
+*/
+function amountOfRowsThatFit() {
+
```
jasonkofo: Pls remove the space

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	from = parseInt(fromValue);
+	let rowsThatFit = amountOfRowsThatFit();
+	to = from + rowsThatFit;
+
+	if (to > 100000) {
+		window.alert("Nee, stout >.<,your search has exceeded the maximum rows")
+		from = 100000 - rowsThatFit;
+		to = 100000;
+	} else if (from < 0) {
+		window.alert("Nee, stout >.<,your search has exceeded the minimum rows")
```
jasonkofo: A few points here: 
- Pls shorten your error message.
- I would rather you not use alerts to get this message across, because it tends to be quite jarring to the user experience of an app
- If you are going to print the same configuration of messages (the principle of not writing the same line of code multiple times), pls extract this error message to a function.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value = from.toString();
+	toValue = (<HTMLInputElement>document.getElementById("toValue")).value = to.toString();
+	callHeaders(from, to);
+
+}
+/**
+ *Executes when the user presses the next button
+ */
+function pressedNext() {
+
```
jasonkofo: pls remove unnecessary spaces

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	let toValue = (<HTMLInputElement>document.getElementById("toValue")).value;
+	let fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value;
+
+	let rowsThatFit = amountOfRowsThatFit();
```
jasonkofo: Rather remove the declaration of this variable and replace it with the snippet that I will attach in the very next comment.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+
+	from = parseInt(toValue) + 1;
+	to = from + rowsThatFit;
```
jasonkofo, line 73:
```ts
to = from + amountOfRowsThatFit();
```
Also, pls shorten the name of the function _amountOfRowsThatFit_.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	if (to > 100000) {
+		window.alert("Nee, stout >.<,your search has exceeded the maximum rows")
+		from = 100000 - rowsThatFit;
+
+		to = 100000;
+	}
+	fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value = from.toString();
+	toValue = (<HTMLInputElement>document.getElementById("toValue")).value = to.toString();
+
+
```
jasonkofo: Pls remove spaces.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	toValue = (<HTMLInputElement>document.getElementById("toValue")).value = to.toString();
+	callHeaders(from, to);
+
+}
+
+
+/**
+ *Retrieves column names
+*/
+function callHeaders(from: number, to: number) {
```
jasonkofo: This method is called _callHeaders_ and the next one is called _retrieveRows_.
Could you pls make the names of these functions a little bit more consistent throughout your code?

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	fetch('http://localhost:2050/columns')
+		.then((resp) => {
+			return resp.json();
+		})
+		.then((columnsRequest) => {
+			let headerData = <string[]>columnsRequest;
+			create(headerData);
```
jasonkofo, line 113: Pls add a (only slightly longer but) more descriptive name for this method.

display.css
```diff
@@ -0,0 +1,53 @@
+
+th{
```
jasonkofo, line 2: For all these styles, could you pls put spaces between the elements and the braces, for example:
```css
td {
 ...
}
```

app.ts
```diff
@@ -0,0 +1,187 @@
...
+	let tableRetrieved = <HTMLInputElement>document.getElementById("intialTable");
+	if (tableRetrieved != null) {
+		tableRetrieved.remove();
+	}
+	let table = document.createElement('table');
+	table.setAttribute('id', 'intialTable');
+	document.getElementById("body")!.appendChild(table);
+
+	let newRow = document.createElement('tr');
+	newRow.setAttribute('id', 'rowNumber0');
```
jasonkofo: could you set the ids in all of your HTML elements to be kebab case (eg. 'row-number-1'), not because this adds much value to THIS onboarding, but because it will add value to the future code that you do write.

app.ts
```diff
@@ -0,0 +1,187 @@
...
+			for (let j = 1; j < 15; j++) {
+				let newRow: HTMLElement = document.createElement('tr');
+				newRow.setAttribute('id', 'rowNumber' + j);
+				document.getElementById("intialTable")!.appendChild(newRow);
+				console.log("new row");
+
+				for (let i = 0; i < 11; i++) {
+					console.log("here");
+					let columnName = document.createElement('td');
+					console.log("evolving");
```
jasonkofo: stray console logs, when finished please remove

jasonkofo (changes requested):
I might comment in a super blunt tone, but it is with the fondest and kindest of intentions

app.ts
nkosimarz: "Typescript disallows your from editing the value late" not really

app.ts
```diff
@@ -0,0 +1,170 @@
...
+	toValue = (<HTMLInputElement>document.getElementById("toValue")).value = to.toString();
+	retrieveHeaders(from, to);
+}
+/**
+ *Adjust the values in the start and end boxes according to the screen heigth whenever it resizes
+ */
+function fromAndToValuesAdjustment() {
+	if (timeout) clearTimeout(timeout);
+	timeout = setTimeout(() => {
+		let toValue = (<HTMLInputElement>document.getElementById("toValue")).value;
```
Winston-Guess, line 46: This should be part of a separate function to reduce nesting IMO

app.ts
```diff
@@ -0,0 +1,170 @@
...
+		retrieveHeaders(from, to);
+	});
+}
+/**
+ *Executes when the user presses the next button
+ */
+function pressedNext() {
+	let toValue = (<HTMLInputElement>document.getElementById("toValue")).value;
+	let fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value;
+	let rowsThatFit = eligableRowCount();
```
Winston-Guess, line 71: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+ *Adjust the values in the start and end boxes according to the screen heigth whenever it resizes
+ */
+function fromAndToValuesAdjustment() {
+	if (timeout) clearTimeout(timeout);
+	timeout = setTimeout(() => {
+		let toValue = (<HTMLInputElement>document.getElementById("toValue")).value;
+		let fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value;
+
+		from = parseInt(fromValue);
+		let rowsThatFit = eligableRowCount();
```
Winston-Guess, line 50: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+/**
+*These are the starting values for when the screen loads
+*/
+function onLoadValues() {
+	let toValue = (<HTMLInputElement>document.getElementById("toValue")).value;
+	let fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value;
+
+	from = 0;
+	fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value = from.toString();
+	let rowsFittable = eligableRowCount();
```
Winston-Guess, line 35: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+	retrieveHeaders(from, to);
+}
+/**
+ *Executes when the user presses the previous button
+ */
+function pressedPrevious() {
+
+	let toValue = (<HTMLInputElement>document.getElementById("toValue")).value;
+	let fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value;
+	let rowsThatFit = eligableRowCount();
```
Winston-Guess, line 90: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+	}
+	fromValue = (<HTMLInputElement>document.getElementById("fromValue")).value = from.toString();
+	toValue = (<HTMLInputElement>document.getElementById("toValue")).value = to.toString();
+	retrieveHeaders(from, to);
+}
+/**
+ *Executes when the user presses the previous button
+ */
+function pressedPrevious() {
+
```
Winston-Guess, line 87: empty line

app.ts
```diff
@@ -0,0 +1,170 @@
...
+/**
+ *Retrieves column names
+*/
+function retrieveHeaders(from: number, to: number) {
+	fetch('http://localhost:2050/columns')
+		.then((resp) => {
+			return resp.json();
+		})
+		.then((columnsRequest) => {
+			let headerData = <string[]>columnsRequest;
```
Winston-Guess, line 112: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+		.then((resp) => {
+
+			return resp.json();
+		})
+		.then((rowsRequest) => {
+
+			let columnData = <string[]>rowsRequest;
+
+			columnData.forEach(function (item) {
+				let newRow: HTMLElement = document.createElement('tr');
```
Winston-Guess, line 134: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+				newRow.setAttribute('id', 'row-number-' + j);
+				document.getElementById("intialTable")!.appendChild(newRow);
+				let content = Array.from(item);
```
Winston-Guess, line 137: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+				content.forEach(function (index) {
+					let columnName = document.createElement('td');
```
Winston-Guess, line 139: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+					indexNumber++;
+				});
+				j++;
+			});
+		});
+}
+
+function create(input: string[]) {
+	//create table
+	let tableRetrieved = <HTMLInputElement>document.getElementById("intialTable");
```
Winston-Guess, line 152: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+	if (tableRetrieved != null) {
+		tableRetrieved.remove();
+	}
+	const table = document.createElement('table');
+	table.setAttribute('id', 'intialTable');
+	document.getElementById("body")!.appendChild(table);
+
+	let newRow = document.createElement('tr');
```
Winston-Guess, line 160: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+	newRow.setAttribute('id', 'row-number-0');
+	document.getElementById("intialTable")!.appendChild(newRow);
+
+	for (let i = 0; i < input.length; i++) {
+		let columnName = document.createElement('th');
```
Winston-Guess, line 165: `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+let from = 0;
+let to = 0;
+
+/**callHeaders
+ * calculates the amount of rows that can fit on the screen given its current height
+*/
+function eligableRowCount() {
+	let windowHeight = window.innerHeight;
+	let buttonDivHeight = (<HTMLInputElement>document.getElementById("buttons")).offsetHeight;
+	let adaptedHeight = windowHeight - buttonDivHeight;
```
Winston-Guess, line 22: all `const`

app.ts
```diff
@@ -0,0 +1,170 @@
...
+	let j = 1;
+
+	fetch('http://localhost:2050/records?from=' + from + '&to=' + to)
+		.then((resp) => {
+
+			return resp.json();
+		})
+		.then((rowsRequest) => {
+
+			let columnData = <string[]>rowsRequest;
```
Winston-Guess, line 131: `const`

Winston-Guess (commented):
Just be careful whenever you use `let`. A `let` signifies to whoever is reading your code that the variable is going to be reassigned later. Always err on the side of using `const`.

Winston-Guess (approved):
Other than those last few minor changes I'm happy. Good work! 👍

PR #33, opened 2020-01-10
Onboarding submission

pierrehenrinortje:
This is the first submission for the onboarding javascript task.
All specifications and requirements have been met.


app.ts
```diff
@@ -0,0 +1,216 @@
...
+    };
+
+    $("#canvas").append(fromField,searchbtn, prevbtn, nextbtn);
+
+    initiate();
+
+    /**
+     * Initiates the HTTP requests to obtain the records and column values necessary for the table.
+     */
+    function initiate() {
```
jasonkofo: I'm going to raise a few issues that I have with this paradiagm you have stuck to. These gripes also apply to the _generateTable_ method.
Firstly, you defined this function using the function tag inside of a callback. There are 2 popular ways to define methods in Javascript which I will share below.
The first is using the function tag e.g.
```ts
    function randomFunctionName(...randomArgs) {
        // Implementation
    }
``` 
The second uses the arrow function declaration paradigm, e.g.
```ts
    randomFunctionName = (...randomArgs) => {
        // Implementation
    }
```
There are a few benefits to using the latter, most of which are vastly beyond the scope this javascript onboarding. However, I believe that the style guide requires you to use arrow functions, rather than the archaic function keyword.
You also declared these functions after you used them, as a matter of conventions (ala what I remember from the Typescript style guide) that isn't desirable
Finally, at least for now and from what I can gather, you only use this function once. Is it entirely necessary to declare this as a method or can we rather use something like a comment to demarcate specific logic?

app.ts
```diff
@@ -0,0 +1,216 @@
+window.onload = () => {
```
jasonkofo, line 69: I have a mild issue with the entirety of your logic living in this onload event handler. Especially considering that you declare another event callback (namely _window.onresize_) within this callback.

My issue here is primarily one of separation of concerns, you want all these sections of your implementation to interact with both the DOM and other sections of your source code in a coherent and cohesive fashion - for the sake of clarity, some other interests, but primarily for the developer that comes after you.

However to change this would require you to think about how you will manage your other data fields like _numColumns_, _columns_ etc. I know that in varsity they taught us to shy away from global variables, but don't be afraid, considering you are here embracing procedural programming :)

Also remember Fritz's complaint about the spacing in your IDE being changed to use tab sizes rather than spaces.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    let prevbtn = document.createElement("button");
+    prevbtn.innerText = "Previous Page";
+
+    let fromField = document.createElement("input");
+    fromField.placeholder = "from";
+
+    // Window resizing (using a debouncing method):
+    window.onresize = () => {
+        let time = 100;
```
jasonkofo: _time_ doesn't look like it changes in this function. As a matter of convention at IMQS, we declare variables using the _const_ keyword instead of _let_ in order to let the developer know that the variable isn't going to be re-assigned (except in the case of arrays, because for some reason Typescript doesn't care whether an array is declared with const or let, but there you go.

Internally, it does not make much of a difference, because all of these consts and lets become vars in javascript upon transpilation, but I know that it is important to some people (ahem, ahem, @FritzOnFire ).

app.ts
```diff
@@ -0,0 +1,216 @@
...
+    nextbtn.innerText = "Next Page";
+
+    let prevbtn = document.createElement("button");
+    prevbtn.innerText = "Previous Page";
+
+    let fromField = document.createElement("input");
+    fromField.placeholder = "from";
+
+    // Window resizing (using a debouncing method):
+    window.onresize = () => {
```
jasonkofo: Also, please do not declare this callback while you're declaring the onload callback, rather move it out into a fresh scope. Declaring the other onclick handlers for the HTML elements are fine, but declaring this new callback on the same object feels wrong to me.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+        }
+    };
+
+    nextbtn.onclick = function() {
+        let end = tableRecordCount - 1;
+        if (recordCount > startIndex + tableRecordCount) {
+            startIndex += tableRecordCount;
+            if (recordCount < startIndex + tableRecordCount) {
+                startIndex = recordCount - tableRecordCount;
+            }
```
jasonkofo: I would prefer you using a ternary operation for this e.g.
```ts
    startIndex = someConditional ? <true assign> : <false assign>
```

app.ts
```diff
@@ -0,0 +1,216 @@
...
+            startIndex = from;
+            if (recordCount < startIndex + tableRecordCount) {
+                startIndex = recordCount - tableRecordCount;
+            }
+            getRecords(startIndex, startIndex + end);
+        }
+    };
+
+    nextbtn.onclick = function() {
+        let end = tableRecordCount - 1;
```
jasonkofo: Considering that you only use this as an offset right at the end (when you call _getRecords_), is it entirely necessary to declare this as a variable of its own?

app.ts
```diff
@@ -0,0 +1,216 @@
...
+        let dataSize = getSize(data);
+        let table = document.createElement("table");
+        let tbdy = document.createElement("tbody");
+        table.appendChild(generateHeadings(columns));
+        for (let i = 0; i < tableRecordCount; i++){
+            let tr = document.createElement("tr");
+            for (let j = 0; j < numColumns; j++){
+                let td = document.createElement("td");
+                if (dataSize > i)
+                    td.appendChild(document.createTextNode(data[i][j]));
```
jasonkofo: I have performance concerns about N^2 operations like this, especially because I know that there is some bottleneck with working with Javascript's native DOM API that I am not entirely familiar with - leading me to conclude this to perhaps be a slightly expensive call.
I know that in the historic IMQS codebase, we use a lot of string generation in order to minimize the number of raw appends that we do (i.e. generate the raw HTML in code and then only make 1 dom call), and that is how it was done in the past.
If you want to stick with the native JS API, I attached a link that speaks about documentfragments.
https://coderwall.com/p/o9ws2g/why-you-should-always-append-dom-elements-using-documentfragments

app.ts
```diff
@@ -0,0 +1,216 @@
...
+        }
+        thead.appendChild(tr);
+        return thead;
+    }
+
+    /**
+     * Validates that the field value is a number.
+     * @param from The ID value from which to start searching.
+     */
+    function validate(from: number) {
```
jasonkofo: I appreciate the existence of this method, but what does this function return?

app.ts
```diff
@@ -0,0 +1,216 @@
...
+        }
+        table.appendChild(tbdy);
+        canvas.appendChild(table);
+    }
+
+    /**
+     * Finds the number of entries within a data set.
+     * @param data An array of data, type is unknown (any).
+     */
+    function getSize(data: any) {
```
jasonkofo: Your method signature seems to return a number, so I would advise that you append the return type to the method signature and leverage the type safety that Typescript makes available.
```ts
    function getSize(data: any): number {
        // Implementation
    }
```
Also, it's not great practice to make your function accept any values, because then we are not leveraging the type safety that TS makes available to us.

package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "onboard-javascript",
+  "version": "1.0.0",
+  "description": "Onboarding",
+  "main": "index.js",
+  "dependencies": {
+    "jquery": "^3.4.1"
```
jasonkofo, line 7: Pls include typescript as an npm package, rather than with apt. The idea is that someone should be able to clone this repo, run _npm install_ and have everything they need, rather than running into the error when they try to run tsc.
```
npm install --save-dev typescript 
``
The _--save-dev_ script saves an entry for the version and for the name of the package.

File: app.ts [hunk 11]
```diff
@@ -0,0 +1,216 @@
...
+        thead.appendChild(tr);
+        return thead;
+    }
+
+    /**
+     * Validates that the field value is a number.
+     * @param from The ID value from which to start searching.
+     */
+    function validate(from: number) {
+        if (isNaN(from)) {
```
jasonkofo: You are leaving your null/undefined handling (in this method) to whatever comes out of the IsNaN method.

Please add a case with the undefined or null instances.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * Adjust the number of records shown in the table according to the window size.
+     */
+    function adjustTableRecordCount() {
+        let height = window.innerHeight;
+        let fontSize = getComputedStyle(document.documentElement).fontSize + "";
+        let rowHeight = parseFloat(fontSize)*2.5;
+        if (rowHeight !== undefined){
```
jasonkofo: rather do something like
```ts 
    if (typeof rowHeight !== 'undefined') {
        // Implementation
    }
```
But as a secondary aside, @FritzOnFire will probably end up telling you to invert the conditional so that you can get rid of the unnecessary indentation in the code by doing something like
```ts
    if (typeof rowHeight === 'undefined') {
        return;
    }
``` 
and then continuing with the implementation after this single conditional.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+            if (recordCount < startIndex + tableRecordCount) {
+                startIndex = recordCount - tableRecordCount;
+            }
+            getRecords(startIndex, startIndex + end);
+        }, time);
+    }
+    
+    // on clicks:
+    searchbtn.onclick = function() {
+        let from = parseInt(fromField.value, 10);
```
jasonkofo: Behold, an extract from the es5 library mthod for Typescript

```ts
/**
 * Converts a string to an integer.
 * @param s A string to convert into a number.
 * @param radix A value between 2 and 36 that specifies the base of the number in numString.
 * If this argument is not supplied, strings with a prefix of '0x' are considered hexadecimal.
 * All other strings are considered decimal.
 */
declare function parseInt(s: string, radix?: number): number;
```
I include this signature to illustrate that adding the 10 at the end is unnecessary. Please remove.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * Generates a table populated with data, distrubuted evenly.
+     * @param data The data records retrieved from the server, structured as a 2D object array.
+     * @param columns An object array containing the column heading values.
+     */
+    function generateTable(data: any, columns: any) {
+        $("table").remove(); // Remove previous table
+        let dataSize = getSize(data);
```
jasonkofo: dataSize and perhaps a few other properties here don't change, redeclare as const.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+        });
+    }
+
+    /**
+     * Generates a table populated with data, distrubuted evenly.
+     * @param data The data records retrieved from the server, structured as a 2D object array.
+     * @param columns An object array containing the column heading values.
+     */
+    function generateTable(data: any, columns: any) {
+        $("table").remove(); // Remove previous table
```
jasonkofo: Why remove the entire table and add everything back, when one need only remove the contents of the table prior to recreating the contents?
In this context, it doesn't mean much, because the function is only called once, but as a matter of practice, it could prove problematic.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * Finds the number of entries within a data set.
+     * @param data An array of data, type is unknown (any).
+     */
+    function getSize(data: any) {
+        let i = 0;
+        let entry;
+        for (entry in data) {
```
jasonkofo: Rather use 
```ts
    for (let entry in data) {
```

app.ts
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * Adjust the number of records shown in the table according to the window size.
+     */
+    function adjustTableRecordCount() {
+        let height = window.innerHeight;
+        let fontSize = getComputedStyle(document.documentElement).fontSize + "";
+        let rowHeight = parseFloat(fontSize)*2.5;
+        if (rowHeight !== undefined){
+            tableRecordCount = Math.trunc(height/rowHeight)-2;
+            if (tableRecordCount < 1)
```
jasonkofo: consider changing to a ternary operator for concise writing.

app.ts
```diff
@@ -0,0 +1,216 @@
...
+     */
+    function validate(from: number) {
+        if (isNaN(from)) {
+            alert("\"From\" field does not have a number value.");
+            return false;
+        } else if (from < 0) {
+            alert("\"From\" value cannot be negative.");
+            return false;
+        }  else if (from > recordCount-1) {
+            alert("\"From\" value exceeds the record count.");
```
jasonkofo: For these text fields, you could have delimited the string using single quotes and then used the double quotes inside the string, to aid the readability of this line
```ts
alert('"From" value exceeds the record count.');
```

jasonkofo (changes requested):
So I've put a lot of comments calling for changes, but I am really pleased about the following:
- Formatting is looking good
- Logic is fairly easy to follow because the methods are well named, and compact in terms of functionality
- Your functions have documentation describing their purposes, I really appreciate these

So, I'll look once again once you've made these changes, but I think that it is looking good.

PR #34, opened 2021-03-09
Onboarding Project

JeanNicholson:


app.css
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#tablediv
+{ 
+    position: absolute;
+    top: 25px;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    height: calc(100% - 50px);
```
FritzOnFire: We prefer to not use calc if we can avoid it.

app.css
```diff
@@ -0,0 +1,162 @@
+.maindiv
```
FritzOnFire: We prefer to keep our `{` on the same line as the class or id for css.

app.ts
```diff
@@ -0,0 +1,293 @@
+class table
+{
+
+  constructor()
```
FritzOnFire: Please convert the formatting to tabs

app.css
```diff
@@ -0,0 +1,162 @@
...
+  {
+    top:0;
+    height: calc(100% - 25px);
+  }
+}*/
+
+@media screen and (max-width:800px) {
+  #next10
+  {
+    display: none !important;
```
jasonkofo: You're using so many `!important`  operators. While I understand that there must have been a solid reason for that, this is not necessarily best practice in css. Generally you can get the result you are looking for by messing with the specificity of your CSS selector

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception

app.ts
```diff
@@ -0,0 +1,293 @@
+class table
+{
```
FritzOnFire: Please go look at the https://github.com/IMQS/www-lib/tree/master/_codestyle#naming-conventions code example and try to stick to those conventions. Also, please put the `{` on the same line as the class declaration.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  constructor()
+  {
+    this.from = 0;
+    this._to = this.to;
+  }
+
+  private _from:number = 0;
+  private _to:number;
```
FritzOnFire: Please move these before the constructor

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  get from()
+  {
```
FritzOnFire: Please put the `{` on the same line as the function declaration and add a return type to this function.

app.ts
```diff
@@ -0,0 +1,293 @@
+class table
```
jasonkofo: I think Fritz is going to mention something about the style guide, but please adhere to it.

On a stylistic point, in Typescript we have chosen to have our classes named using PascalCase - but I am sure that the style guide makes mention of something along those lines

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  private _from:number = 0;
+  private _to:number;
+
+  get from()
+  {
+    return this._from;
+  }
+
+  set from(from:number)
```
FritzOnFire: Your formatter seems to not be kicking in. there should be a space between the `:` and the type

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  { 
+    let totalRecordsIndex:number = getTotalRecords() - 1;
+
+    console.log("test");
+    console.log(this._from);
+    console.log(totalRecordsIndex);
+    console.log(this.getNumberOfRows());
```
FritzOnFire: Please remove all stray console.logs unless they are genuine errors

app.ts
jasonkofo: This is a comment that is going to persist onto every single opening brace that we will see in the codebase. I am not going to mention it because it is going to surround you with endless noise.

But let's write these opening braces using cuddled bracing - i.e. the brace is on the same line as the prototype, where possible

https://blog.gskinner.com/archives/2008/11/curly_braces_to.html

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
```
jasonkofo: I'm sure at this point we could potentially remove these logs

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    this._to = this.to;
+
+  }
+
+  get to():number
+  {
+    let to:number;
+
+    to = this._from + this.getNumberOfRows()
```
FritzOnFire: You should create and set the variable on the same line.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+
+  //determine the amount of rows to add to the table based on the size on the window
+  getNumberOfRows()
+  { 
+    let rows:number;
+
+    //subtract - 1 to cater for header row.
+    let height = (parseInt(((window.innerHeight-75)/30).toFixed(0)) - 1);
+    let width = (parseInt(((window.innerWidth-75)/30).toFixed(0)) - 1);
```
FritzOnFire: Im not sure why you are taking the width into consideration when you are calculating the amount of rows you need to fetch.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    console.log(this.getNumberOfRows());
+
+    if((from + this.getNumberOfRows()) > totalRecordsIndex)
+      this._from = (totalRecordsIndex - this.getNumberOfRows());
+    else if(from < 0 )
+      this._from = 0;
+    else
+      this._from = from;
+
+    this._to = this.to;
```
jasonkofo: So the internal gets re-assigned to the value that is given at to?
From what I am seeing (please correct me if I am wrong), there is no real need for to `_to` variable if it exists for the purpose of being re-written once the `from` cached value is re-calculated.

Unless there is something else here, can we please remove it?

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  constructor()
+  {
+    this.from = 0;
+    this._to = this.to;
+  }
+
+  private _from:number = 0;
+  private _to:number;
+
+  get from()
```
jasonkofo: type this function

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    else if(from < 0 )
+      this._from = 0;
+    else
+      this._from = from;
+
+    this._to = this.to;
+
+  }
+
+  get to():number
```
jasonkofo: type this function

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    if (to > getTotalRecords())
+      return getTotalRecords() - 1;
+    else
+      return to;
+  }
+
+
+  //determine the amount of rows to add to the table based on the size on the window
+  getNumberOfRows()
```
jasonkofo: type this function

app.ts
```diff
@@ -0,0 +1,293 @@
...
+      rows = width;
+
+    if (rows < 0)
+      return 0;
+    else
+      return rows; 
+  
+  }
+
+  getTotalRecords()
```
jasonkofo: type this function

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    let width = (parseInt(((window.innerWidth-75)/30).toFixed(0)) - 1);
+
+    if (height < width)
+      rows = height;
+    else
+      rows = width;
+
+    if (rows < 0)
+      return 0;
+    else
```
jasonkofo: This last else is not necessary

app.ts
```diff
@@ -0,0 +1,293 @@
...
+      return rows; 
+  
+  }
+
+  getTotalRecords()
+  {
+    const HttpRequest = new XMLHttpRequest();
+    const url='http://localhost:2050//recordCount';
```
FritzOnFire: I dont think you need two `/` in front of `recordCount`

app.ts
```diff
@@ -0,0 +1,293 @@
...
+      dataType:'json',
+      success:function(data) {
+
+        $.each(data, function(row)
+          {
+            const dataRow = tableBody.insertRow(-1);
+            $.each(data[row], function(cell)
+              {
+                const newCell = dataRow.insertCell(-1);
+                newCell.innerHTML = data[row][cell];
```
jasonkofo: My recommendation would be that this internal logic in the for loop needs to be re-imagined to build up HTML as a string and then write the HTML to the DOM using 1 read operation, rather than `10*(num_records_fetched)` DOM read operations.

It's not entirely necessary, it just seems a tad bit wasteful.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  getTotalRecords()
+  {
+    const HttpRequest = new XMLHttpRequest();
+    const url='http://localhost:2050//recordCount';
+    HttpRequest.open("GET", url, false);
+    HttpRequest.send();
+
+    const responseText = HttpRequest.responseText
```
FritzOnFire: The send method is an asynchronous method. Which means, that if you server responds to slow, then you will be setting `null` into your `responseText`. You definitely want to handle this differently.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    const dataRow = tableBody.insertRow(0);
+
+    //make the ajax call to retrieve the records and build the table header
+    $.ajax({
+      url: `http://localhost:2050/columns`,
+      dataType:'json',
+      async:false,
+      success:function(data) {
+
```
jasonkofo: empty lines that can be removed

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  { 
+    $("#dataTableBody").find("tr").remove();
+
+    let tableBody = <HTMLTableSectionElement> document.getElementById("dataTableBody");  
+
+    //make the ajax call to retrieve the records and build the table
+    $.ajax({
+      url: `http://localhost:2050/records?from=${from}&to=${to}`,
+      dataType:'json',
+      success:function(data) {
```
FritzOnFire: You should use an arrow function instead, eg. `() => {}`

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+        $.each(data, function(row)
```
FritzOnFire: Please change this to a `for..of` loop https://github.com/IMQS/www-lib/tree/master/_codestyle#loops

app.ts
```diff
@@ -0,0 +1,293 @@
...
+          {
+            const dataRow = tableBody.insertRow(-1);
```
FritzOnFire: instead of constantly editing a dom element that is being rendered, you should rather create a new table and only add it to the dom when you are done.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    });
+
+    //call the button property set fucntion to set disable/enably buttons appropriately 
+    buttonPropertySet();
+  
+  }
+
+}
+
+class utility
```
jasonkofo: I have a few comments here:
- PascalCase for a class name.
- `utility` is somewhat of a vague name for a class. What is this thing supposed to do? If it has a single debounce utility, and if my next comment didn't exist, I would have suggested that a better name be `DebounceUtility` or something like that
- Making a class whose only purpose it to expose debounce functionality seems a bit excessive to me - because to my eyes, you are essentially just making a class to receive the exact number of arguments that the setTimeout method would expect anyways, which only really managing the ID of the timeout object, and call a baked version of the function. 
I remember mentioning something about a singleton class that you could use to manage state, but now that you have gone the class route for this, this debounce functionality ought to be something that your class itself manages, because of separation of concerns. This means that the properties and some debounce logic are moved into the `table` (soon to be `Table`) class with the addition of a method for managing the timeout ID, rather than as a separate class.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+      const later = () => {
+        clearTimeout(this.timeout);
+        func(...args);
+      };
+  
+      clearTimeout(this.timeout);
+      this.timeout = setTimeout(later, wait);
+    };
+  };
+}
```
jasonkofo: I'd much rather see this code structured in this sort of manner:
```ts
                clearTimeout(this.debounceProp);
		this.debounceProp = setTimeout(() => {
			func(...args);
		}, this.timeout);
```

This, of course being in the newer implementation where this debounce logic is moved into the `Table` class

app.ts
```diff
@@ -0,0 +1,293 @@
...
+        //const dataRow = tableBody.insertRow(0);
+        $.each(data, function(cell)
+          {
+            const newCell = dataRow.insertCell(-1);
+            newCell.outerHTML = `<th>${data[cell]}</th>`;
+             
+          }
+        )
+      }
+    });
```
FritzOnFire: Pretty sure you only need to call this once, and not everytime you need to build the table.

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+}
+
+//on resize funtionalty to rebuild the table
+$(window).on('resize', function() 
+{
+  // wrap the logic within a debounce funtion to prevent unnecesary calls. 
+  let request = myUtility.debounce(function(){
+
+    myTable.from = myTable.from;
```
jasonkofo: The only possible explanation that I have for this line is that you are trying to force Typescript to recompute this getter - is this true?

app.ts
```diff
@@ -0,0 +1,293 @@
...
+window.onload = function()
+{
+
+  const request = myUtility.debounce(myTable.buildTable,250);
+
+  request(myTable.from,myTable.to);
+
+}
+
+//on resize funtionalty to rebuild the table
```
FritzOnFire: Please start you comment with a space

app.ts
FritzOnFire: Not sure what is happening here

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    next5.disabled=false;
+    next10.disabled=false;
+  }
+}
+
+//function to retrieve the total record count used when building the table
+function getTotalRecords()
+{
+    const HttpRequest = new XMLHttpRequest();
+    const url='http://localhost:2050//recordCount';
```
jasonkofo: Does the string need to be escaped or can this just be
```suggestion
    const url='http://localhost:2050/recordCount';
```

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  else
+  {
+    next.disabled=false;
+    next5.disabled=false;
+    next10.disabled=false;
+  }
+}
+
+//function to retrieve the total record count used when building the table
+function getTotalRecords()
```
jasonkofo: please type this function

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  }
+  else
+  {
+    next.disabled=false;
+    next5.disabled=false;
+    next10.disabled=false;
+  }
+}
+
+//function to retrieve the total record count used when building the table
```
FritzOnFire: Please use javadoc style comments for functions. `/** comment */`

app.ts
```diff
@@ -0,0 +1,293 @@
...
+function getTotalRecords()
+{
+    const HttpRequest = new XMLHttpRequest();
+    const url='http://localhost:2050//recordCount';
+    HttpRequest.open("GET", url, false);
+    HttpRequest.send();
+
+    const responseText = HttpRequest.responseText
```
FritzOnFire: Same comment as before about the send method

app.ts
```diff
@@ -0,0 +1,293 @@
...
+  }
+  else
+  {
+    previous.disabled=false;
+    previous5.disabled=false;
+    previous10.disabled=false;
+  }
+
+  // disable next buttons when out of range
+  if(from + myTable.getNumberOfRows() === (totalRecords -1))
```
jasonkofo: Interested to know why you are checking for hard equality and not just if the value that is returned is greater than the total records? Do you have another condition for handling this case?

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+//on resize funtionalty to rebuild the table
+$(window).on('resize', function() 
+{
+  // wrap the logic within a debounce funtion to prevent unnecesary calls. 
+  let request = myUtility.debounce(function(){
+
+    myTable.from = myTable.from;
+
+    myTable.buildTable(myTable.from,myTable.to);
```
FritzOnFire: Why are you passing myTable variables that it all ready has?

index.html
```diff
@@ -1,12 +1,28 @@
...
+		<script type="text/javascript" charset="utf-8" src="app.js"></script>
+		<link href="app.css" rel="stylesheet" />
+	</head>
+
 
 <body>
-	<p>Hello</p>
+	<div class="maindiv">
+		<div id="captiondiv">Data Grid</div>
+		<div id="tablediv"><table id="dataTable"><tbody id="dataTableBody"></tbody></table></div>
```
jasonkofo: We generally use kebab-casing for our HTML IDs and classes, rather than camelCasing

index.html
```diff
@@ -1,12 +1,28 @@
...
+		<div id="buttondiv">
+      
+			<button onclick="previousButton(10)" id="previous10" class="buttonClass">&lt&lt&lt 10 </button>
+			<button onclick="previousButton(5)" id="previous5" class="buttonClass">&lt&lt 5</button>
+			<button onclick="previousButton(1)" id="previous" class="buttonClass">&lt</button>
+			<button onclick="nextButton(1)" id="next" class="buttonClass">&gt</button>
+			<button onclick="nextButton(5)" id="next5" class="buttonClass">5 &gt&gt</button>
+			<button onclick="nextButton(10)" id="next10" class="buttonClass">10 &gt&gt&gt</button>
+      <button onclick="jumpToButton()" class="buttonClass" id="jumpToButton">Jump To:</button>
```
jasonkofo: Same comment about kebab-casing as above

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+//previous button function that takes a multiplier indicating the amount of pages to page at a time
+function previousButton(multiplier:number) 
+{
+  // wrap the logic within a debounce funtion to prevent unnecesary calls. 
+  let request = myUtility.debounce(function(){
+  
+    myTable.from = myTable.from - ((myTable.getNumberOfRows() + 1) * multiplier);
+
+    myTable.buildTable(myTable.from,myTable.to);
```
FritzOnFire: Why are you passing `myTable` variables that is all ready has?

app.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    //make the ajax call to retrieve the records and build the table header
+    $.ajax({
+      url: `http://localhost:2050/columns`,
+      dataType:'json',
+      async:false,
+      success:function(data) {
+  
+        //const dataRow = tableBody.insertRow(0);
+        $.each(data, function(cell)
```
jasonkofo: arrow function
```suggestion
        $.each(data, (cell: <type>) =>
```

app.ts
```diff
@@ -0,0 +1,293 @@
...
+    tableBody = <HTMLTableSectionElement> document.getElementById("dataTableBody");
+
+    const dataRow = tableBody.insertRow(0);
+
+    //make the ajax call to retrieve the records and build the table header
+    $.ajax({
+      url: `http://localhost:2050/columns`,
+      dataType:'json',
+      async:false,
+      success:function(data) {
```
jasonkofo:
```suggestion
      success: (data: <type>) => {
```

app.ts
jasonkofo: arrow function
```suggestion
      success: (data: <type>) => {
```

index.html
```diff
@@ -1,12 +1,28 @@
...
+      <input type="number" id="jumpToValue" value="0"></input>
+		</div>
+	</div>
```
FritzOnFire: I really think you should make these in code.

FritzOnFire (changes requested):
Now, there is nothing wrong with having more than one class in the same file, but you have a lot of other logic mixed in with your classes, and app.ts is clearly where you app starts and not where you should be defining classes. Please try and create some more files

app.ts
```diff
@@ -0,0 +1,298 @@
...
+    return parseInt(responseText);
+  }
+
+  buildTable(from:number,to:number) 
+  { 
+    $("#dataTableBody").find("tr").remove();
+
+    let tableBody = <HTMLTableSectionElement> document.getElementById("dataTableBody");
+
+    console.log("test speed 1");
```
jasonkofo: stray console log

app.ts
```diff
@@ -0,0 +1,298 @@
...
+    return (...args:any[]) => {
+      const later = () => {
+        clearTimeout(this.timeout);
+        func(...args);
+      };
+  
+      clearTimeout(this.timeout);
+      this.timeout = setTimeout(later, wait);
+    };
+  };
```
jasonkofo: I would prefer this function to look something like
```suggestion
      clearTimeout(this.timeout);
      this.timeout = setTimeout(() => {
          func(...args);
      }, wait);
    };
  };
```
This of course is once the functionality has been moved into the `table` (soon to be `Table`) class

This of course, being when it is not returning the callback, but rather executing the function on demand.

jasonkofo (changes requested):
I got a bit tired, especially because I am sure @FritzOnFire is going to cover the other style issues in his PR review. I chose to simply overlook those issues, because it doesn't seem that this code was developed in accordance with the style guide - which is okay until the next sweep of the PR. 

A lot of the comments I made are suggestions about the general code design. Please feel free to ask me what I meant if some of the comments that I might give off an air of insanity, or pedantry, because that unfortunately is a thing that often happens.

app.ts
```diff
@@ -0,0 +1,143 @@
...
+	let from = myTable.from;
+	let totalRecords = myTable.totalRecords;
+
+	// disable previous buttons when out of range
+	if (from === 0) {
+		previous.disabled = true;
+		previous5.disabled = true;
+		previous10.disabled = true;
+	}
+	else {
```
FritzOnFire, line 34: Please put this on the same line.

table.ts
```diff
@@ -0,0 +1,138 @@
...
+			this._from = (totalRecordsIndex - this.rows);
+		else if (from < 0)
+			this._from = 0;
+		else
+			this._from = from;
+
+	}
+
+	get to(): number {
+		let to = this._from + this.getNumberOfRows()
```
FritzOnFire, line 41: please put a `;` at the end of this line.

table.ts
```diff
@@ -0,0 +1,138 @@
...
+
+			// make the ajax call to retrieve the records and build the table
+			$.ajax({
+				url: `http://localhost:2050/records?from=${this.from}&to=${this.to}`,
+				dataType: 'json',
+				success: (data: string[]): void => {
+
+					for (let row of data) {
+
+						html = html + "<tr>"
```
FritzOnFire, line 119: this can be `html += "<tr>";`

table.ts
```diff
@@ -0,0 +1,138 @@
...
+
+						for (let cell of row) {
+							html = html + "<td>" + cell + "</td>";
```
FritzOnFire, line 122: this can be `html += "<td>" + cell + "</td>";`

table.ts
```diff
@@ -0,0 +1,138 @@
...
+						}
+
+						html = html + "</tr>"
```
FritzOnFire, line 125: this can be `html += "</tr>";`

FritzOnFire, line 125: this can be `html += "</tr>";`

FritzOnFire (commented):
Some formatting comments, but you are done in my opinion :D

PR #38, opened 2021-06-21
Gerald - Onboarding

geraldtivatyi:


FritzOnFire:
Please remove your `.js` and `.js.map` files

geraldtivatyi:
> Please remove your `.js` and `.js.map` files

I removed the files and changed gitignore file.

app.ts
```diff
@@ -0,0 +1,309 @@
+import { TableHeadingString } from "./classes/table-headings.js";
+import { RenderTableRows } from "./classes/render-rows.js";
+import { RenderTableHeading } from "./classes/render-headings.js";
+import { HasFormatMethod } from "./interfaces/has-format-method.js";
```
FritzOnFire: You should not reference `.js` files. You should be able to reference them without any extension.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+let toID = numOfRows;
+fromIDElement.innerHTML = fromID.toString();
+toIDElement.innerHTML = toID.toString();
+
+// Create initial page with initial data
+createInitialPage(numOfRows);
+
+// Listen for change in window size
+let screenHeight = screen.height;
+window.addEventListener('resize', function(event) {
```
FritzOnFire: You should change this to an arrow function

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		numOfRows = 23;
+	} else if ((screenHeight < 750) && (screenHeight >= 550)) {
+		numOfRows = 18;
+	} else if ((screenHeight < 550) && (screenHeight >= 265)) {
+		numOfRows = 8;
+	} else if ((screenHeight < 265) && (screenHeight >= 95)) {
+		numOfRows = 3;
+	} else if (screenHeight < 95) {
+		numOfRows = 2;
+	}
```
FritzOnFire: Please rework this code to be more dynamic. This is feels very much like a hard coded solution.

app.ts
He7i: I feel like there should be a debounce in here somewhere,fritz?

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		toID = numOfRows;
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
+	} else {											// Else, continue creating table with the set IDs
+		fromID = fromID - (numOfRows-1);
+		toID = toID - (numOfRows-1);
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
+
+		// Create table using the new set IDs inside a set timeout
```
He7i: From here to line 95 gets repeated frequently,consider moving this functionality into a seperate method

app.ts
```diff
@@ -0,0 +1,309 @@
...
+					
+// Listen for click on right arrow
+$( "#rightarrow" ).on( "click", function() {
+	// Clear the timeout every time if you navigate and you're not at the end ID of all the data
+	if (toID != totalNumofRecords) {
+			clearTimeout(clickTimeout);
+	}
+
+	// Clear form value in front-end when navigating data using the arrows
+	let startfrom = document.querySelector("#startfrom") as HTMLInputElement;
```
He7i: I also feel these two lines get repeated a lot and can be moved into an independent method

app.ts
```diff
@@ -0,0 +1,309 @@
...
+
+	// Recreate grid with new set number of rows
+	createInitialPage(numOfRows);
+});
+
+// Set a timeout to use for promises when navigating through data
+let clickTimeout: ReturnType<typeof setTimeout>;
+
+// When clicking on left arrow
+$( "#leftarrow" ).on( "click", function() {
```
FritzOnFire: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	// Clear the timeout every time if you navigate and you're not at the start ID of all the data
+	if (fromID != 1) {
+		clearTimeout(clickTimeout);
+	}
+
+	// Clear form value in front-end when navigating data using the arrows
+	let startfrom = document.querySelector("#startfrom") as HTMLInputElement;
+	startfrom.value = "";
+	
+	if ((fromID < (numOfRows-1)) && (fromID > 1)) {		// If the "from" ID is within the IDs at the beginning of the dataset
```
FritzOnFire: You should remove all the extra braces.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		fromID = 1;
+		toID = numOfRows;
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
+
+		// Create table using the new set IDs inside a set timeout
+		clickTimeout = setTimeout(function(){
```
FritzOnFire: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,309 @@
...
+			// Use resolved promises to fetch data from backend
+			let headings = Promise.resolve(fetch("/columns").then(res => res.text()));
```
FritzOnFire: You should only fetch the headings once in your application.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		// Create table using the new set IDs inside a set timeout
+		clickTimeout = setTimeout(function(){
+			// Use resolved promises to fetch data from backend
+			let headings = Promise.resolve(fetch("/columns").then(res => res.text())); 
+			let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
+			// Promise.all<string>([headings, records]).then((values) => {
+			Promise.all<string>([headings, records]).then((values) => {
+				generateTable(values[0], values[1]);
+			}).catch(err => console.log(err));
+		}, 500);
```
FritzOnFire: You want to keep this number as small as possible. With 250 being the absolute max.

app.ts
FritzOnFire: Pretty sure you should be using `.json()`

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		fromID = 1;
+		toID = numOfRows;
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
+
+		// Create table using the new set IDs inside a set timeout
+		clickTimeout = setTimeout(function(){
+			// Use resolved promises to fetch data from backend
+			let headings = Promise.resolve(fetch("/columns").then(res => res.text())); 
+			let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
```
FritzOnFire: Pretty sure you should be using `.json()`

app.ts
FritzOnFire: You also don't need `Promise.resolve`

app.ts
FritzOnFire: You also don't need `Promise.resolve`

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		// Create table using the new set IDs inside a set timeout
+		clickTimeout = setTimeout(function(){
+			// Use resolved promises to fetch data from backend
+			let headings = Promise.resolve(fetch("/columns").then(res => res.text()));
+			let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
+			Promise.all<string>([headings, records]).then((values) => {
+				generateTable(values[0], values[1]);
+			}).catch(err => console.log(err));
+		}, 500);
+	}
```
FritzOnFire: I think you should rework this function to only handle the `fromID` and `toID` in the `if` and `else if`s. And then probably a boolean to decide if you are going to kick off a request to the back-end.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+});
+					
+// Listen for click on right arrow
+$( "#rightarrow" ).on( "click", function() {
```
FritzOnFire: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		clickTimeout = setTimeout(function(){
+			// Use resolved promises to fetch data from backend
+			let headings = Promise.resolve(fetch("/columns").then(res => res.text())); 
+			let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
+			Promise.all<string>([headings, records]).then((values) => {
+				generateTable(values[0], values[1]);
+			}).catch(err => console.log(err));
+		}, 500);
+	}
+});
```
FritzOnFire: Your functions for `#rightarrow` and `#leftarrow` are kinda similar. I would like you to rework them to ONLY change the `fromID` and `toID` and then have them both call a method that handles the debounceing, updating of the front-end and the fetching of the rows.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+							
+// Listen for submission in form and use inputs to request data from backend
+$( "#submit" ).on( "click", function() {
```
FritzOnFire: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	let startFromIDElement = document.querySelector('#startfrom') as HTMLInputElement;
+	let startFrom = startFromIDElement.valueAsNumber;
+
+	// Checks to see if you request invalid values in the form
+	if (startFrom < 1 || startFrom > (totalNumofRecords-(numOfRows-1))) {
+		alert("The acceptable range is between 1 and "+(totalNumofRecords-(numOfRows-1)));
+		return;
+	} else if (isNaN(startFrom)) {
+		alert("You have not set a value to submit.");
+	} else {
```
FritzOnFire: You probably want to return after that alert.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		fromID = startFrom;
+		toID = fromID + (numOfRows-1);
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
```
FritzOnFire: This code should not be in an else as your previous two `if` statements both return.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+
+	// Create table using the new set IDs inside a set timeout
+	clickTimeout = setTimeout(function(){
+		// Use resolved promises to fetch data from backend
+		let headings = Promise.resolve(fetch("/columns").then(res => res.text())); 
+		let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
+		Promise.all<string>([headings, records]).then((values) => {
+			generateTable(values[0], values[1]);
+		}).catch(err => console.log(err));
+	}, 500);
```
FritzOnFire: This code should be replace with the same function that you call at the end of `#rightarrow` and `#leftarrow`

app.ts
```diff
@@ -0,0 +1,309 @@
...
```
FritzOnFire: This code should be replace with the same function that you call at the end of `#rightarrow` and `#leftarrow`

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	
+	// Create table using the new set IDs inside a set timeout
+	clickTimeout = setTimeout(function() {
+		// Use resolved promises to fetch data from backend
+		let headings = Promise.resolve(fetch("/columns").then(res => res.text())); 
+		let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
+		Promise.all<string>([headings, records]).then((values) => {
+			generateTable(values[0], values[1]);
+		}).catch(err => console.log(err));
+	}, 500);
```
FritzOnFire: This code should be replace with the same function that you call at the end of `#rightarrow` and `#leftarrow`

app.ts
```diff
@@ -0,0 +1,309 @@
...
+
+	// Recreate table with new headings and records data
+	let interfaceHeading: HasFormatMethod;							// variable of type interface used in creating table headings
+	let interfaceRecords: HasFormatMethod;							// variable of type interface used in creating table rows
+	interfaceHeading = new TableHeadingString(headingsStr);			// call method to generate string containing table heading element
+	tble.constructTableHeadings(interfaceHeading);					// call method to render table headings element in browser
+	interfaceRecords = new RenderTableRows(recordsStr);				// call method to generate string containing table row elements
+}
+
+// Function to target an element in html by its id
```
FritzOnFire: Please change this comment to JavaDoc style.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+function el(s: string) {
```
FritzOnFire: There have been alot of opportunities to use this function in the code before this function. Why did you not use it there?

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	return document.getElementById(s);
+}
+
+// Function to create the initial table when loading the page for the first time or when the window size changes
```
FritzOnFire: Please change this comment to JavaDoc style.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+function createInitialPage(numOfRows: number) {
+	// Clear the timeout every time this function is called
+	clearTimeout(clickTimeout);
+
+	// Check if you're at the start or end of the data set, and reset IDs for change in window size
+	if (fromID > (totalNumofRecords-(numOfRows-1))) {
+		fromID = totalNumofRecords-(numOfRows-1);
+		toID = fromID + (numOfRows-1);
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
```
FritzOnFire: These 3 rows seem to be exactly the same in both your `if` and `else`. You should move them outside.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	//Request for and set the total number of records
+	fetch("/recordCount").then(res => res.text()).then((value) => {
+		numofrecords.innerHTML = value;
+		totalNumofRecords = Number(value) - 1;
+	}).catch(err => console.log(err))
```
FritzOnFire: Don't you need this information to set the `toID`? What would happen if your `toID` was bigger than `totalNumofRecords`?

FritzOnFire: Your also missing a `;` at the end of your line.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+
+	// Create table using the new set IDs inside a set timeout
+	clickTimeout = setTimeout(function(){
+		// Use resolved promises to fetch data from backend
+		let headings = Promise.resolve(fetch("/columns").then(res => res.text())); 
+		let records = Promise.resolve(fetch("/records?from="+fromID+"&to="+toID).then(res => res.text()));
+		Promise.all<string>([headings, records]).then((values) => {
+			generateTable(values[0], values[1]);
+		}).catch(err => console.log(err));
+	}, 500);
```
FritzOnFire: This code should be replace with the same function that you call at the end of `#rightarrow` and `#leftarrow`

app.ts
FritzOnFire: You should rather just call `getNumOfRows` and rework that function rather.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	} else if ((screenHeight < 550) && (screenHeight >= 265)) {
+		numOfRows = 8;
+	} else if ((screenHeight < 265) && (screenHeight >= 95)) {
+		numOfRows = 3;
+	} else if (screenHeight < 95) {
+		numOfRows = 2;
+	}
+
+	// Recreate grid with new set number of rows
+	createInitialPage(numOfRows);
```
FritzOnFire: I don't think your resize function should be calling `createInitialPage`. You should just be figuring out which rows to fetch, fetch them and then rebuild just the part of the table that needs to be changed.

index.html
```diff
@@ -1,13 +1,37 @@
...
+					<p>of</p>
+					<p id="numofrecords" name="numofrecords"></p>
+				</div>
+				<svg id="rightarrow" class="arrow" width="24" height="24" xmlns="http://www.w3.org/2000/svg" fill-rule="evenodd" clip-rule="evenodd"><path d="M12 0c6.623 0 12 5.377 12 12s-5.377 12-12 12-12-5.377-12-12 5.377-12 12-12zm0 1c6.071 0 11 4.929 11 11s-4.929 11-11 11-11-4.929-11-11 4.929-11 11-11zm-3 5.753l6.44 5.247-6.44 5.263.678.737 7.322-6-7.335-6-.665.753z"/></svg>
+			</div>
+		</div>
+		<div id="app">
+			<div id="table" class="table"></div>
+		</div>
+		<script type="module" charset="utf-8" src="app.js"></script>
```
FritzOnFire: You should keep your scripts in your `head` tag.

interfaces/has-format-method.ts
```diff
@@ -0,0 +1,5 @@
+// interface to ensure methods implementing this interface contain a internalFormat method that returns a string (the string generated for the elements inside the table)
```
FritzOnFire: Please change this comment to JavaDoc style

FritzOnFire (changes requested):
You have a lot of variables and statements executing in global space which is generally frown'd upon. For this project, there are a few that we are ok with being in global space, but you are running a lot of them without waiting for the page to even finish loading.

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	} else if (fromID === 1) {							// If the "from" ID is 1 then set the values and do not make any backend calls
+		fromID = 1;
+		toID = numOfRows;
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
+	} else {											// Else, continue creating table with the set IDs
+		fromID = fromID - (numOfRows-1);
+		toID = toID - (numOfRows-1);
+		fromIDElement.innerHTML = fromID.toString();
+		toIDElement.innerHTML = toID.toString();
```
nkosimarz: these lines can be outside the if statement and dont have to be repeated in each if/else
		fromIDElement.innerHTML = fromID.toString();
		toIDElement.innerHTML = toID.toString();

app.ts
geraldtivatyi: It actually should be .text() due to how I coded my methods for handling the data from the backend calls

FritzOnFire (changes requested):
Please read up on window.onload. All of this global space statements are really bothering me. Please move them somewhere more sane. Please change all of your comments for functions and classes to JavaDoc style. Please do not use an interface before you have defined it (HasFormatMethod) and finally, it really looks like your formatter is not kicking in. I will send you my vscode settings incase you can not get your formatting working.

app.ts
```diff
@@ -0,0 +1,276 @@
+// Global Variables
+let screenHeight = screen.height;								
+let numOfRows = getNumOfRows();		// Call function to get number of rows to render based on current screen size
+let totalNumofRecords: number;						
+let clickTimeout: ReturnType<typeof setTimeout>;	
+let fromID = 1;										
+let toID = numOfRows;
```
nkosimarz: I dont think all of these need to be global, especially numofrows

nkosimarz (changes requested):

app.ts
```diff
@@ -0,0 +1,354 @@
...
+/// <reference path="./classes/render-table-headings.ts" />
+/// <reference path="./classes/render-table-rows.ts" />
+/// <reference path="./classes/get-column-headings.ts" />
+
+// Global variable
+let timeOut: ReturnType<typeof setTimeout>;
+
+/**
+   * This event executes two funtions immediately after the page loads
+   * @function HFM.getColumnHeadings This is a function that gets the column headings from the back-end
```
nkosimarz, line 12: I prefer this pattern for comments: @function HFM.getColumnHeadings gets the column heading ...

nkosimarz (approved):
The structure is definitely cleaner, LGTM

app.ts
geraldtivatyi: Thanks. I'll quickly change the comments to your suggestion

PR #40, opened 2021-08-19
updated onboarding project

SAMMOLOI:


app.ts
```diff
@@ -0,0 +1,374 @@
+namespace onboardproject {
+
```
nkosimarz: remove empty lines

app.ts
```diff
@@ -0,0 +1,374 @@
...
+    let Load = 50,previous: number[],previousprocess: number;
+
+    // region API Call 
+   // Trigger async function
+    async function getRecordCountCall() :  Promise<number> {
+      
+        const response = await fetch('http://localhost:2050/recordCount');
+        
+        let promise = new Promise((res, rej) => {
+            setTimeout(() => res("Now it's done!"), 50)
```
nkosimarz: is this meant to be Load

app.ts
```diff
@@ -0,0 +1,374 @@
...
+    getRecordCountCall().then(data => console.log(data)).catch(reason => console.log(reason.message))
+
+    // Trigger async function
+    async function getColumnNamesCall() : Promise<string[]>{
+
+        const response = await fetch('http://localhost:2050/columns');
+
+        let promise = new Promise((res, rej) => {
+
+            setTimeout(() => res("Now it's done!"),50)
```
nkosimarz: is this meant to be Load

app.ts
```diff
@@ -0,0 +1,374 @@
...
+        throw new Error("Error");
+        
+    }
+
+    function calculateToId(fromId: number): number {
+
+        const possibleRecordsData = Math.floor((window.innerHeight - ($("#form-content").innerHeight() as number)) / 37);
+        const possibleId = fromId + possibleRecordsData;
+
+        let recordDisplayset = 0;
```
nkosimarz: isn't this always zero just before you switch on it? al lot of hardcoding happening here can you add a comment to elaborate what you are trying to accomplish here

nkosimarz (changes requested):
You need to get a formatter for your VScode and also get rid of a lot of empty line

app.ts
SAMMOLOI: I am manage to fix my format and empty line on the code

app.ts
SAMMOLOI: Yes, which mean my async function is going to be pause until a Promise is Settled (that is, fulfilled or rejected), and to resume execution of the async function after fulfillment( i was using it ,to test my function inside )

app.ts
SAMMOLOI: Yes, which mean my async function is going to be pause until a Promise is Settled (that is, fulfilled or rejected), and to resume execution of the async function after fulfillment( i was using it ,to test my function inside )

app.ts
```diff
@@ -0,0 +1,238 @@
+namespace onboardproject {
+	module onboardprojects {
+		//Variable declarations
+		let Load = 50, previous: number[], previousprocess: number;
+		// region API Call
```
FritzOnFire: Please have an empty line between your variable declarations and the start of your functions declarations.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+		// Trigger async function
+		async function getRecordCountCall(): Promise<number> {
+			const response = await fetch('http://localhost:2050/recordCount');
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				throw new Error(message);
+				console.log(message)
```
FritzOnFire: You should not have code following a `throw`

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			} else {
+				return await response.json();
+				console.log(response);
```
FritzOnFire: You should not have code following a `return`

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			}
+		}
+		// log response or catch error of fetch promise
+		getRecordCountCall().then(data => console.log(data)).catch(reason => console.log(reason.message))
```
FritzOnFire: Please move all global function calls into your `window.onload` function. (Or anywhere where it is not in global space)

app.ts
```diff
@@ -0,0 +1,238 @@
...
+
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				throw new Error(message);
+			} else {
+				return await response.json();
+				console.log(response);
```
FritzOnFire: You should not have any code following a `return`

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			}
+		}
+		// trigger async function
+		// log response or catch error of fetch promise
+		getColumnNamesCall().then(data => console.log(data)).catch(reason => console.log(reason.message))
```
FritzOnFire: Same comment as before about global functions.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			const response = await fetch(`http://localhost:2050/records?from=${(fromID)}&to=${(toID)}`);
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				throw new Error(message);
+			} else {
+				return await response.json();
+				console.log(response);
```
FritzOnFire: You should not have any code following a `return`

app.ts
```diff
@@ -0,0 +1,238 @@
...
+		//     } else {
+		//         // We reached our target server, but it returned an error
+
+		//         console.log("Error 404");
+		//     }
+		// };
+		// }
+		//region Data Loading methods
+		async function LoadRecordsData(fromID: number, toID: number): Promise<number[]> {
+			try {
```
FritzOnFire: Why do you need the `try-catch`? The function calling this function should be able to catch the error of the Promise that you are returning.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+				const recordsvalue = await getRecordsCall(fromID, toID)
```

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					DisplayContent += `<tr id="table-row-${record[0]}">`;
+					for (const column of record) {
+						DisplayContent += `<td align="center">${column}</td>`;
+					}
+					DisplayContent += '</tr>';
+					$("#wrapper-table-content-body").empty();
+					$("#wrapper-table-content-body").append(DisplayContent);
+				}
+			} catch (error) {
+				console.log("Error <No data>" + error)
```
FritzOnFire: You should rather add a catch to line 77 and either log or throw there.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			}
+			return [fromID, toID];
```
FritzOnFire: Why return what the caller gave him?

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			cursor = cursor.sort((a, b) => { return a - b });
+			return LoadRecordsData(cursor[0], cursor[1]);
+			// throw new Error("Error");
+		}
+		window.onresize = () => {
+			try {
+				const nextToId = calculateToId(previous[0]);
+				clearTimeout(Load);
+				Load = setTimeout(async () => {
+					const recordCount = await getRecordCountCall();
```
FritzOnFire: You are allowed to skip this step. For this project the record count will not change after the initial call.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					if (nextToId >= recordCount - 1) {
+						const fromId = recordCount - 1 - (calculateToId(previous[0]) - previous[0]);
+						const toId = recordCount - 1;
+						previous = await LoadRecordsData(fromId, toId);
+					} else {
+						previous = await LoadRecordsData(previous[0], nextToId)
```

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					}
+				}, 10);
```
FritzOnFire: 10 is quite quick... You should change it to 250. Anything around the human response time. Or at least, what is considered to be the average.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					if (parseAsInt == true) {
+						return parseInt(input as string);
+					}
+					return parseFloat(input as string);
+				case ("number"):
+					return input as number;
+				default:
+					return 0;
+			}
+			throw new Error("Error");
```
FritzOnFire: Because of the default case in your switch statement this line will never hit. If you must have a return outside of your switch, then make it the same as your default.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+				case 1:
+					window.innerHeight <= 969;
+					break;
+				case 2:
+					window.innerHeight <= 1938;
+					break;
+				default:
+					recordDisplayset = 15;
+					break;
+			}
```
FritzOnFire: The result of this bit of code will always have recordDisplayset set to 0... im not sure what your trying to achive here. but the hardcoding for handling diffrent window heights is not what you should be doing. Why is `possibleId` on its own not enough?

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			const fromID = ConvertNumber(previous.sort((a, b) => { return a - b })[0]);
+			const toID = ConvertNumber(previous.sort((a, b) => { return a - b })[1]);
+			const documentHeight = $(window).innerHeight() as number - ($(`#table-row-${fromID}`).height() as number);
+			for (let i = fromID; i <= toID; i++) {
+				const elementHeightOffset = ($(`#table-row-${i}`).offset() as JQueryCoordinates).top;
+				if (elementHeightOffset < documentHeight) continue;
+				return i;
+			}
+			return toID;
+		}
```
FritzOnFire: This function is cool, but very expensive. Rather do some math like you did in `calculateToId`.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+		function previousPageResize(previous: number[]): number[] {
+			const toId = calculateToId(previous[0] - (nextPageResize(previous) - previous[0]));
+			return [previous[0] - (nextPageResize(previous) - previous[0]), toId];
+		}
+		window.onload = async () => {
+			try {
+				previous = await LoadPageContent(0, calculateToId(0));
```
FritzOnFire: Rather add a `.catch` to this line.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+				$("#previous").click(async () => {
+					const CountData = await getRecordCountCall();
```
FritzOnFire: You are allowed to skip this step. For this project the record count will not change after the initial call.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					let toId = (previous[0] >= 0 ? previous[1] : possibleStep);
+					fromId = fromId == CountData - 1 ? fromId - possibleStep : fromId;
+					toId = toId <= CountData - 1 ? toId : CountData - 1;
+					previous = await LoadRecordsData(fromId, toId);
+				});
+			} catch (error) {
+			}
+			$("#next").click(async () => {
+				try {
+					const recordCount = await getRecordCountCall();
```
FritzOnFire, line 158: You are allowed to skip this step. For this project the record count will not change after the initial call.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					const possibleStep = calculateToId(fromId) - fromId;
+					if (fromId <= recordCount - possibleStep - 1) {
+						const toId = fromId + possibleStep <= recordCount - 1 ? fromId + possibleStep : recordCount - 1;
+						previous = await LoadRecordsData(fromId, toId);
+					} else if (fromId <= recordCount - 1) {
+
+						previous = await LoadRecordsData(recordCount - 1 - (calculateToId(fromId) - fromId), recordCount - 1);
+					} else {
+						console.log("Test is working ");
+					}
```
FritzOnFire: You should remove debugging `console.log`s

app.ts
```diff
@@ -0,0 +1,238 @@
...
+			} catch (error) {
+			}
+			$("#next").click(async () => {
+				try {
+					const recordCount = await getRecordCountCall();
+					const fromId = nextPageResize(previous);
+					const possibleStep = calculateToId(fromId) - fromId;
+					if (fromId <= recordCount - possibleStep - 1) {
+						const toId = fromId + possibleStep <= recordCount - 1 ? fromId + possibleStep : recordCount - 1;
+						previous = await LoadRecordsData(fromId, toId);
```
FritzOnFire: Rather add a `.catch`

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					} else if (fromId <= recordCount - 1) {
+
+						previous = await LoadRecordsData(recordCount - 1 - (calculateToId(fromId) - fromId), recordCount - 1);
```
FritzOnFire: Rather add a `.catch`

app.ts
```diff
@@ -0,0 +1,238 @@
...
+					} else {
+						console.log("Test is working ");
+					}
+				} catch (error) {
+				}
+			});
+			$("#go-button").click(async () => {
+				try {
+					const recordCount = await getRecordCountCall();
```
FritzOnFire: You are allowed to skip this step. For this project the record count will not change after the initial call.

app.ts
```diff
@@ -0,0 +1,238 @@
...
+						if (Math.floor(fromId).toString() == fromId.toString() === true) {
+							if (fromId > recordCount - possibleStep) {
+								alert(`You may not insert a desired Id greater than ${recordCount - possibleStep}`);
+							} else {
+								let toId = (fromId) + possibleStep < recordCount ? (fromId) + possibleStep : recordCount - 1;
+								previous = await LoadRecordsData(fromId, toId);
+							}
+						} else {
+							console.log("Test is working ");
+						}
```
FritzOnFire: You should remove debugging `console.log`s

index.html
```diff
@@ -2,11 +2,32 @@
...
 	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+	<script src="app.js"> </script>
+	<link rel="stylesheet" type="text/css" href="/style.css">
 </head>
 
-<body>
-	<p>Hello</p>
+<body id="body">
+	<div id="form-content">
+			Go To Start Index:<input type="number" id="index" value="0">
```
FritzOnFire: I think you have one indent to many here.

FritzOnFire: You should add a `min` here. to prevent the user from inputting negative numbers.

index.html
```diff
@@ -2,11 +2,32 @@
...
+	</div>
+	<table id='wrapper-table'>
+		<thead id="wrapper-table-header-row">
+			
+		</thead>
+		<tbody id="wrapper-table-content-body">
+
+		</tbody>
+		</table>
+	<div id="table-arbitrary-footer" ></div></div>
```
FritzOnFire: I think you have one to many closing div tags

index.html
```diff
@@ -2,11 +2,32 @@
...
+
+	
+
+
```
FritzOnFire: You should remove these empty lines

style.css
```diff
@@ -0,0 +1,91 @@
+
```
FritzOnFire: Remove this empty line.

style.css
```diff
@@ -0,0 +1,91 @@
+	
+	table tr td,table tr td{
```
FritzOnFire: Please check the formatting on this file... something seems to have gone wrong.

FritzOnFire (changes requested):
In general, remove your `try-catches`.  All of them seem to be in functions that are async and return promises so the calling function should be able to catch any errors.

Please put an empty line between your function definitions.

FritzOnFire:
Also, why is your previouses an array? I think you should rather just have a number or two.

app.ts
```diff
@@ -0,0 +1,241 @@
...
+					DisplayContent += `<tr id="table-row-${record[0]}">`;
+					for (const column of record) {
+						DisplayContent += `<td align="center">${column}</td>`;
+					}
+					DisplayContent += '</tr>';
+					$("#wrapper-table-content-body").empty();
+					$("#wrapper-table-content-body").append(DisplayContent);
+				}
+			} catch (error) {
+				console.log("Error <No data>" + error)
```
pierrehenrinortje: missing `;`

style.css
SAMMOLOI: I manage to fix the duplication.
	this code is was selecting two titles that are the same.table tr td,table tr td{

app.ts
```diff
@@ -0,0 +1,243 @@
...
+					}
+					return parseFloat(input as string);
+				case ("number"):
+					return input as number;
+				default:
+					return 0;
+			}
+			throw new Error("Error");
+		}
+		function calculateToId(fromId: number): number {
```
pierrehenrinortje: Add a concise comment above the function, this is importent when the purpose of a function is not completely obvious.

app.ts
```diff
@@ -0,0 +1,243 @@
...
+		// Trigger async function
+		async function getRecordCountCall(): Promise<number> {
+			const response = await fetch('http://localhost:2050/recordCount');
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				//throw new Error(message);
+				console.log(message)
```
pierrehenrinortje: missing `;`

app.ts
```diff
@@ -0,0 +1,243 @@
...
+					}
+				} catch (error) {
+				}
+			});
+			$("#go-button").click(async () => {
+				const recordCount = await getRecordCountCall();
+				const fromId = ConvertNumber($("#index").val() as string, false);
+				const possibleStep = calculateToId(fromId) - fromId;
+				if (fromId < 0) {
+					alert('only insert Id greater than or equal to 0');
```
pierrehenrinortje, line 178: add a return here so we can remove the next "else" below.  This is more readable and reduces excessive indentation in cases where larger nested "if else" statements exist.

app.ts
```diff
@@ -0,0 +1,243 @@
...
+				} else {
+					if (Math.floor(fromId).toString() == fromId.toString() === true) {
+						if (fromId > recordCount - possibleStep) {
+							alert(`You may not insert a desired Id greater than ${recordCount - possibleStep}`);
+						} else {
+							let toId = (fromId) + possibleStep < recordCount ? (fromId) + possibleStep : recordCount - 1;
+							previous = await LoadRecordsData(fromId, toId);
+						}
+					} else {
+						console.log("Test is working ");
```
pierrehenrinortje: remove this debug logging

pierrehenrinortje (changes requested):
Agree with Fritz, some extra comments on my end.

app.ts
```diff
@@ -0,0 +1,236 @@
+import { data } from "jquery";
+
+namespace onboardproject {
+	module onboardprojects {
+
+		//Load Variable declarations
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,236 @@
...
+		let Load = 50;
+		// Previous Variable declarations
+		let previous: number[];
+		//Previous Process Variable declarations
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,236 @@
...
+		// region API Call
+		// Trigger async function
+		async function getRecordCountCall(): Promise<number> {
+			const response = await fetch('http://localhost:2050/recordCount');
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				//throw new Error(message);
```
FritzOnFire: Why did you remove the `throw`?

app.ts
```diff
@@ -0,0 +1,236 @@
...
+				console.log(message)
+			} else {
+				console.log(response);
+			}
+			return await response.json();
```
FritzOnFire: You should not await on a promise you are returning.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+		async function getColumnNamesCall(): Promise<string[]> {
+			const response = await fetch('http://localhost:2050/columns');
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				throw new Error(message);
+			} else {
+				return await response.json();
```
FritzOnFire: You should not await on a promise you are returning.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+		async function getRecordsCall(fromID: number, toID: number): Promise<string[][]> {
+			const response = await fetch(`http://localhost:2050/records?from=${(fromID)}&to=${(toID)}`);
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
+			if (!response.ok) {
+				const message = `An error has occured: ${response.status}`;
+				throw new Error(message);
+			} else {
+				return await response.json();
```
FritzOnFire: You should not await on a promise you are returning.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+				console.log(response);
+			}
+		}
+		// function requestcolumns<Request>(
```
FritzOnFire: Please add an empty line above this line.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+		//         const data = JSON.parse(this.response) as Response;
+		//         callback && callback(data);
+		//     } else {
+		//         // We reached our target server, but it returned an error
+
+		//         console.log("Error 404");
+		//     }
+		// };
+		// }
+		//region Data Loading methods
```
FritzOnFire: Please add a space after the `//`, and an empty line above this line.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+				for (const column of record) {
+					DisplayContent += `<td align="center">${column}</td>`;
+				}
+				DisplayContent += '</tr>';
+				$("#wrapper-table-content-body").empty();
+				$("#wrapper-table-content-body").append(DisplayContent);
+			}
+			return [fromID, toID];
+		}
+		function RecordsFromCursor(cursor: number[]): Promise<number[]> {
```
FritzOnFire: Please add an empty line above this line.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+			cursor = cursor.sort((a, b) => { return a - b });
+			return LoadRecordsData(cursor[0], cursor[1]);
+			// throw new Error("Error");
+		}
+		window.onresize = () => {
```
FritzOnFire: Please move this into `window.onload`

app.ts
```diff
@@ -0,0 +1,236 @@
...
+		// Handlers
+		async function LoadPageContent(fromID: number, toID: number): Promise<number[]> {
+			let DisplayContent = "";
+			const columns = await getColumnNamesCall();
+			for (const column of columns) {
+				DisplayContent += `<th align="center">${column}</th>`;
+				$("#wrapper-table-header-row").empty();
+				$("#wrapper-table-header-row").append(DisplayContent);
+			}
+			return await LoadRecordsData(fromID, toID);
```
FritzOnFire: You should not await on a promise you are returning.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+			const toID = ConvertNumber(previous.sort((a, b) => { return a - b })[1]);
+			const documentHeight = $(window).innerHeight() as number - ($(`#table-row-${fromID}`).height() as number);
+			for (let i = fromID; i <= toID; i++) {
+				const elementHeightOffset = ($(`#table-row-${i}`).offset() as JQueryCoordinates).top;
+				if (elementHeightOffset < documentHeight) continue;
+				return i;
+			}
+			return toID;
+		}
+		function previousPageResize(previous: number[]): number[] {
```
FritzOnFire: Please add an empty line above this.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+			const toId = calculateToId(previous[0] - (nextPageResize(previous) - previous[0]));
+			return [previous[0] - (nextPageResize(previous) - previous[0]), toId];
+		}
+		window.onload = async () => {
+			// trigger async function
+			// log response or catch error of fetch promise
+			getColumnNamesCall().then(data => console.log(data)).catch(reason => console.log(reason.message));
+			// trigger async function
+			// log response or catch error of fetch promise
+			getRecordCountCall().then(data => console.log(data)).catch(reason => console.log(reason.message));
```
FritzOnFire: These functions look like debugging code. You should remove them if they are.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+						previous = await LoadRecordsData(recordCount - 1 - (calculateToId(fromId) - fromId), recordCount - 1);
+					} else {
+						console.log("Test is working ");
+					}
+				} catch (error) {
+				}
+			});
+			//Searching function for index
+			$("#go-button").click(async () => {
+				const recordCount = await getRecordCountCall();
```
FritzOnFire, line 174: You are allowed to skip this step. For this project the record count will not change after the initial call.

FritzOnFire (changes requested):

app.ts
SAMMOLOI: Yes, I was using it to view, if my function had problems.

app.ts
SAMMOLOI: it was a mistake when I was testing the code. I comment it out

app.ts
SAMMOLOI: the average is 250

SAMMOLOI: An average timeout is 250

app.ts
SAMMOLOI: I started first using function without async and I was getting a lot of errors.i forgot to remove that try-catch on async function after implementing it.

app.ts
SAMMOLOI: The first time when I test the calculateToId Function its was not handling different window heights. that is why I possibleId to determine the window heights.

app.ts
SAMMOLOI: because the promise is that it gives you something to return right away.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+
+		//Previous Process Variable declarations
+		let previousprocess: number;
+
+		// Trigger async function
+		async function getRecordCountCall(): Promise<number> {
+			const response = await fetch('http://localhost:2050/recordCount');
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
```
FritzOnFire: I don't think you need these comments anymore.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+			}
+			return response.json();
+		}
+
+		// Trigger async function
+		async function getColumnNamesCall(): Promise<string[]> {
+			const response = await fetch('http://localhost:2050/columns');
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
```
FritzOnFire: I don't think you need these comments anymore.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+			}
+			return response.json();
+		}
+
+		// Trigger async function
+		async function getRecordsCall(fromID: number, toID: number): Promise<string[][]> {
+			const response = await fetch(`http://localhost:2050/records?from=${(fromID)}&to=${(toID)}`);
+			// let promise = new Promise((res, rej) => {
+			// 	setTimeout(() => res("Now it's done!"), 50)
+			// });
```
FritzOnFire: I don't think you need these comments anymore.

app.ts
FritzOnFire: My comment is still relevant on this line.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+			let DisplayContent = '';
+			for (const record of recordsvalue) {
+				DisplayContent += `<tr id="table-row-${record[0]}">`;
+				for (const column of record) {
+					DisplayContent += `<td align="center">${column}</td>`;
+				}
+				DisplayContent += '</tr>';
+				$("#wrapper-table-content-body").empty();
+				$("#wrapper-table-content-body").append(DisplayContent);
+				throw new Error("Error No data");
```
FritzOnFire: Would this not throw an error every time you call this function?

app.ts
```diff
@@ -0,0 +1,226 @@
+import { data } from "jquery";
+
+namespace onboardproject {
+	module onboardprojects {
+
+		//Load Variable declarations
+		let Load = 50;
+
+		// Previous Variable declarations
+		let previous: number[];
```
FritzOnFire: Please replace this array with one or two variables. I see you do a lot of sorting on this array and only ever use two the the numbers. You really want to avoid all of that.

app.ts
FritzOnFire: My comment on this line is still relevant.

app.ts
FritzOnFire: My comment on this line is still relevant.

app.ts
FritzOnFire: But what your returning is data that he all ready has... so this is just burning CPU time for nothing. Please change the return type to `Promise<void>`.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+				DisplayContent += '</tr>';
+				$("#wrapper-table-content-body").empty();
+				$("#wrapper-table-content-body").append(DisplayContent);
+				throw new Error("Error No data");
+			}
+			return [fromID, toID];
+		}
+
+		// Load Records Function
+		function RecordsFromCursor(cursor: number[]): Promise<number[]> {
```
FritzOnFire: You should rather take in two variables, sorting is not nice, and very expensive when you only care about two of the numbers in the array.

app.ts
FritzOnFire: No, I agree with that, I just don't get why you add the recordDisplayset to the possibleId? And either way, if I removed the lines I commented about then it would have no affect on the result, so I dont know why you have them. You should really remove them.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+					return parseFloat(input as string);
+				case ("number"):
+					return input as number;
+			}
+			return 0;
+		}
+
+		// Height Diplay Function
+		function calculateToId(fromId: number): number {
+			const possibleRecordsData = Math.max((window.innerHeight - ($("#form-content").innerHeight() as number)));
```
FritzOnFire: Why are you getting the max of one number? And the way that you use the result does not make a lot of sens, this looks like it is adding the pixels to the fromId, as if a row is only one pixel tall. I think you accidentally removed a divide somewhere here.

app.ts
```diff
@@ -0,0 +1,236 @@
...
+				case 2:
+					window.innerHeight <= 1938;
+					break;
+				default:
+					recordDisplayset = 15;
+					break;
+			}
+			return recordDisplayset + possibleId;
+		}
+		function nextPageResize(previous: number[]): number {
```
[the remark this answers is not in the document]
FritzOnFire: The comment on this line is still relevant.

app.ts
FritzOnFire: The comment on this function is still VERY VERY relevant.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+			const documentHeight = $(window).innerHeight() as number - ($(`#table-row-${fromID}`).height() as number);
+			for (let i = fromID; i <= toID; i++) {
+				const elementHeightOffset = ($(`#table-row-${i}`).offset() as JQueryCoordinates).top;
+				if (elementHeightOffset < documentHeight) continue;
+				return i;
+			}
+			return toID;
+		}
+
+		//Onload Function
```
FritzOnFire: Please add a space after the `//`, and an empty line above this line.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+		window.onload = async () => {
+
+			//Previous_Page_Resize Function
```
FritzOnFire: Please add a space after the `//`, and an empty line above this line.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+			function previousPageResize(previous: number[]): number[] {
+				const toId = calculateToId(previous[0] - (nextPageResize(previous) - previous[0]));
+				return [previous[0] - (nextPageResize(previous) - previous[0]), toId];
+			}
+
+			//On Resize_Function
```
FritzOnFire: Please add a space after the `//`, and an empty line above this line.

app.ts
```diff
@@ -0,0 +1,226 @@
...
+			return toID;
+		}
+
+		//Onload Function
+		window.onload = async () => {
+
+			//Previous_Page_Resize Function
+			function previousPageResize(previous: number[]): number[] {
+				const toId = calculateToId(previous[0] - (nextPageResize(previous) - previous[0]));
+				return [previous[0] - (nextPageResize(previous) - previous[0]), toId];
```
FritzOnFire: You should rather store the result of `nextPageResize` and use that, than calling that function multiple times

app.ts
FritzOnFire: The comment on this line is still relevant.

FritzOnFire (changes requested):
So the formatting looks a lot better. But for me to approve this, I am going to need you to remove all the `try-catch`s, and you have to stop using an array to store your fromid and toid (it just feels like an anti-pattern).

Oh, and add your `dist/app.min.js` to your gitignore.

PR #41, opened 2021-08-26
Leo On-boarding project

leomoko:
Hi @FritzOnFire 

Can you kindly provide me with a code review

Regards
Leo


app.ts
```diff
@@ -0,0 +1,147 @@
...
+    <button onclick="nextButton()">Next</button>`;
+}
+
+function Search(e) {
+  let inputValue = (<HTMLInputElement>document.getElementById("search")).value;
+
+  if (inputValue === "") {
+    inputValue = startIndex.toString();
+    loadGrid(parseInt(inputValue));
+  } else {
```
pierrehenrinortje: You can remove the else statement here and move `loadGrid(parseInt(inputValue));` outside the if statement.

app.ts
```diff
@@ -0,0 +1,147 @@
+let recordCount: number, startIndex: number, endIndex: number;
+let columnHeaders: any[];
+let resizeTimer: number;
+
+window.onload = () => getRecordCount();
+ 
+//calculating how many records to display based on widow size
```
pierrehenrinortje: Please leave a whitespace after the `//` in comments
This applies to all comments throughout your code.

app.ts
```diff
@@ -0,0 +1,147 @@
...
+    let tableRow = document.createElement("tr");
+    table.appendChild(tableRow);
+
+    let tdArray = records[a].length;
+
+    for (let b = 0; b < tdArray; b++) {
+      tableRow
+        .appendChild(document.createElement("td"))
+        .appendChild(document.createTextNode(records[a][b]));
+
```
pierrehenrinortje: remove this empty line

app.ts
```diff
@@ -0,0 +1,147 @@
...
+  //creating header and search input field
+  document.getElementById("header").innerHTML = `<input type="number" 
+    id="search" onfocusout="Search(event)" placeholder="Search : Enter ID">`;
+
+  //calling Records
+  getRecords(startIndex, endIndex);
+}
+
+function getRecordCount() {
+  //calling record count from the api, full url http://localhost:2050/recordcount
```
pierrehenrinortje: the comment here can be placed above the function declaration

app.ts
```diff
@@ -0,0 +1,147 @@
...
+  $.get("/recordCount", function (numberRecords) {
+    recordCount = numberRecords;
+    getColumnHeaders();
+  }).fail(function () {
+    alert("Error while calling Record Count");
+  });
+}
+
+function getColumnHeaders() {
+  //calling column header from the api, full url http://localhost:2050/columns
```
pierrehenrinortje: the comment here can be placed above the function declaration

app.ts
```diff
@@ -0,0 +1,147 @@
...
+  $.get("/columns", function (headers) {
+    columnHeaders = JSON.parse(headers);
+    loadGrid(0);
+  }).fail(function () {
+    alert("Error while calling column header");
+  });
+}
+
+function getRecords(start: number, end: number) {
+  //calling records from the api, full url http://localhost:2050/records?from=1&to=3
```
pierrehenrinortje: the comment here can be placed above the function declaration

app.ts
```diff
@@ -0,0 +1,147 @@
...
+    endIndex = recordCount - 1;
+  }
+  if (startIndex >= recordCount) {
+    startIndex = recordCount - calRowPerView();
+  }
+  if (startIndex >= recordCount - calRowPerView() && startIndex <= recordCount - 1) {
+    startIndex = recordCount - (calRowPerView() + 1);
+  }
+  //creating header and search input field
+  document.getElementById("header").innerHTML = `<input type="number"
```
pierrehenrinortje: Set the search input field and the buttons in index.html, this way the they won't have to update every time the grid loads data.

app.ts
```diff
@@ -0,0 +1,147 @@
...
+  });
+}
+
+//building table
+function createTable(records: any[]) {
+ //clearing current table
+  document.getElementById("divTable").innerHTML = "";
+  
+ let table = document.createElement("table");
+  // let table = this.$el("table");
```
pierrehenrinortje: remove this commented out code.

app.ts
```diff
@@ -0,0 +1,147 @@
+let recordCount: number, startIndex: number, endIndex: number;
+let columnHeaders: any[];
+let resizeTimer: number;
+
+window.onload = () => getRecordCount();
+ 
+//calculating how many records to display based on widow size
+function calRowPerView() {
```
pierrehenrinortje, line 11: I would make this function set a value outside the function, something like `rowsPerView`. This will reduce the overhead of calling this function every time the value is required. This function is only ever required at least once, or when resizing the window.

pierrehenrinortje (changes requested):
Overall, not bad. I have a few issues I've adressed below that you can attend to.
Another thing, change your indentation from 2 spaces to 4, check that your formatter is inline with our requirements.

app.ts
leomoko: Did you manage to clone repo and get a view of the changes I've done regarding the grid update ?

pierrehenrinortje: I did yes. Alternatively, you could replace `getElementById` with JQuery's `$(#<id>)`, and set the DOM elements this way. But it needs to happen only once.

leomoko: @FritzOnFire did approve the getElementById but i will make the change :+1:

app.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function Search(e) {
+  let inputValue = (<HTMLInputElement>document.getElementById("search")).value;
+
+  if (inputValue === "") {
+    inputValue = startIndex.toString();
+    loadGrid(parseInt(inputValue));
+  } 
+    loadGrid(parseInt(inputValue));
+
```
pierrehenrinortje: remove this empty line

pierrehenrinortje (changes requested):

app.ts
```diff
@@ -0,0 +1,135 @@
...
+  let x = Math.floor((window.innerHeight - 245) / 34);
+  
+  if (x < 0) {
+    alert("Window screen to small to display records , Resize window");
+  }
+  return x;
+}
+
+function loadGrid(start: number) {
+  rowsPerView = calRowPerView();
```
pierrehenrinortje: Here, rowsPerView is assigned a value every time data is loaded, it needs to be assigned only on startup or when the window is resized.

pierrehenrinortje (changes requested):

app.ts
```diff
@@ -0,0 +1,135 @@
...
+  // Adding table to table div ID = divTable
+  document.getElementById("divTable").appendChild(table);
+}
+
+function Search(e) {
+  let inputValue = (<HTMLInputElement>document.getElementById("search")).value;
+
+  if (inputValue === "") {
+    inputValue = startIndex.toString();
+    loadGrid(parseInt(inputValue));
```
nkosimarz, line 110: I don't see any checks for invalid values, the user should get some kind of indicator the input an invalid value rather than auto correction, or at least a hint on what a valid input should look like

nkosimarz (changes requested):

app.ts
```diff
@@ -0,0 +1,137 @@
+let recordCount: number, startIndex: number, endIndex: number;
+let columnHeaders: any[];
+let resizeTimer: number , rowsPerView : number;
```
FritzOnFire, line 3: Formatting is a bit off here. You might want to check that your vscode is formatting your code.

app.ts
```diff
@@ -0,0 +1,137 @@
...
+
+window.onload = function(){
+  getRecordCount();
```
FritzOnFire, line 6: Please use tab's instead of spaces.

app.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (startIndex >= recordCount - rowsPerView && startIndex <= recordCount - 1) {
+    startIndex = recordCount - (rowsPerView + 1);
+  }
+  // calling Records
+  getRecords(startIndex, endIndex);
+}
+
+ // calling record count from the api, full url http://localhost:2050/recordcount
+function getRecordCount() {
+  $.get("/recordCount", function (numberRecords) {
```
FritzOnFire, line 42: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    recordCount = numberRecords;
+    getColumnHeaders();
+  }).fail(function () {
```
FritzOnFire, line 45: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    alert("Error while calling Record Count");
+  });
+}
+
+// calling column header from the api, full url http://localhost:2050/columns
+function getColumnHeaders() {
+  $.get("/columns", function (headers) {
```
FritzOnFire, line 52: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    columnHeaders = JSON.parse(headers);
+    loadGrid(0);
+  }).fail(function () {
```
FritzOnFire, line 55: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    alert("Error while calling column header");
+  });
+}
+
+// calling records from the api, full url http://localhost:2050/records?from=1&to=3
+function getRecords(start: number, end: number) {
+  $.get("/records?from=" + start + "&to=" + end, function (rec) {
```
FritzOnFire, line 62: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    //creating table grid with records returned by the api
+    createTable(JSON.parse(rec));
+  }).fail(function () {
```
FritzOnFire, line 65: Please change this to an arrow function

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    alert("Couldn't load records , page will refresh");
+  });
+}
+
+// building table
+function createTable(records: any[]) {
+ // clearing current table
+  document.getElementById("divTable").innerHTML = "";
```
FritzOnFire, line 73: I recommend you only clean out the table the moment before you set the new one. That will avoid any annoying flashes on slow machines. (Move this line to just before 102)

app.ts
```diff
@@ -0,0 +1,137 @@
...
+  
+  let table = document.createElement("table");
+  let tableBody = document.createElement("tbody");
+  let tableHead = document.createElement("thead");
+  // creating table columns
+  for (let c = 0; c < columnHeaders.length; c++) {
```
FritzOnFire, line 79: You should use `i` (`j` or `k` depending on if this is a nested for loop) for your for loop counters.

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    tableHead
+      .appendChild(document.createElement("th"))
+      .appendChild(document.createTextNode(columnHeaders[c]));
+  }
+  table.appendChild(tableHead);
+
+  // creating table rows
+  for (let a = 0; a < records.length; a++) {
```
FritzOnFire, line 87: Same comment as before about counters

app.ts
```diff
@@ -0,0 +1,137 @@
...
+    let tableRow = document.createElement("tr");
+    table.appendChild(tableRow);
+
+    let tdArray = records[a].length;
+
+    for (let b = 0; b < tdArray; b++) {
```
FritzOnFire, line 93: Same comment as before about counters. But here it should be `j`

app.ts
```diff
@@ -0,0 +1,137 @@
...
+        .appendChild(document.createTextNode(records[a][b]));
+      tableBody.appendChild(tableRow);
+    }
+  }
+  table.appendChild(tableBody);
+  // Adding table to table div ID = divTable
+  document.getElementById("divTable").appendChild(table);
+}
+
+function Search(e) {
```
FritzOnFire, line 105: You should add a type to the parameter

app.ts
```diff
@@ -0,0 +1,137 @@
...
+  let inputValue = (<HTMLInputElement>document.getElementById("search")).value;
+
+  if (inputValue === "") {
+    inputValue = startIndex.toString();
+    loadGrid(parseInt(inputValue));
+  } 
+    loadGrid(parseInt(inputValue));
```
FritzOnFire, line 112: Why do you call this line twice?

app.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// update table gride size based on current window size
+window.onresize = () => {
+  clearTimeout(resizeTimer);
+  resizeTimer = setTimeout(function () {
+    calRowPerView();
+    loadGrid(startIndex);
+  }, 100);
+};
```
FritzOnFire, line 137: I would recommend moving this into the `window.onload` function. In case this runs to soon.

tsconfig.json
```diff
@@ -2,8 +2,9 @@
 	"compilerOptions": {
 		"sourceMap": true,
 		"strict": true,
+		"strictNullChecks": false,
 		"noImplicitThis": true,
-		"noImplicitAny": true,
+		"noImplicitAny": false,
```
FritzOnFire, line 7: You are not allowed to modify this file.

FritzOnFire (changes requested):
Very nice, and very clean, good job :) And if it was not for the formatting and the modifying of the tsconfig.json file I would have said you are done, but Im going to need you to make changes once more.

PR #44, opened 2022-04-12
Devin Fledermaus

DevinFledermaus:
Onboarding Javascript

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -1,6 +1,5 @@
 let fromParameter = 0;
 let recordCount: number;
-const el = document.getElementById
 
 // Gets The 2nd paramater from the 1st parameter
 function getParameters(fromParameter: number): number {
@@ -66,7 +65,7 @@ function getHeadings(): Promise<void> {
 	})
 		.then(handleErrors)
 		.then((response: Response) => response.json())
-		.then((headingData: string | string[]) => {
+		.then((headingData: string[]) => {
 			for (let heading of headingData) {
 				createHeadingColumn(heading);
 			}
@@ -99,7 +98,7 @@ function getTable(): Promise<void> {
 	})
 		.then(handleErrors)
 		.then((res: Response) => res.json())
-		.then((contentData: string | string[]) => {
+		.then((contentData: string[]) => {
 			for (let x of contentData) {
 				appendTableContent(x);
 			}
@@ -153,18 +152,16 @@ function resizing() {
 // Next
 class Next {
 	nextButton: HTMLElement | null;
-	nextTimer: number = 0
+	nextTimer: number = 0;
 
 	constructor() {
 		this.nextButton = document.getElementById("next");
 
 		const nextDebounce = (fn: () => void, delay: number) => {
-			return () => {
-				clearTimeout(this.nextTimer);
-				this.nextTimer = setTimeout(() => {
-					fn();
-				}, delay);
-			};
+			clearTimeout(this.nextTimer);
+			this.nextTimer = setTimeout(() => {
+				fn();
+			}, delay);
 		};
 
 		let next = () => {
@@ -186,7 +183,7 @@ class Next {
 				fromParameter = toParameter - getNoOfRows();
 			}
 
-			nextDebounce(getTable, 500)()
+			nextDebounce(getTable, 500);
 		};
 
 		if (this.nextButton) {
@@ -198,18 +195,16 @@ class Next {
 // Previous
 class Prev {
 	prevButton: HTMLElement | null;
-	prevTimer: number = 0
+	prevTimer: number = 0;
 
 	constructor() {
 		this.prevButton = document.getElementById("prev");
 
 		const prevDebounce = (fn: () => void, delay: number) => {
-			return () => {
-				clearTimeout(this.prevTimer);
-				this.prevTimer = setTimeout(() => {
-					fn();
-				}, delay);
-			};
+			clearTimeout(this.prevTimer);
+			this.prevTimer = setTimeout(() => {
+				fn();
+			}, delay);
 		};
 
 		let prev = () => {
@@ -228,7 +223,7 @@ class Prev {
 					fromParameter = intOne;
 				}
 
-				prevDebounce(getTable, 500)()
+				prevDebounce(getTable, 500);
 			}
 		};
 
@@ -243,11 +238,11 @@ function idJump() {
 	const input: HTMLInputElement | null = document.querySelector("input");
 	let toParameter = getParameters(fromParameter);
 	let currentID = fromParameter;
-	let search: string
+	let search: string;
 	if (input) {
 		search = input.value;
 	} else {
-		return alert("ERROR!!!")
+		return alert("ERROR!!!");
 	}
 	let end = parseInt(search) + getNoOfRows();
 	let maxRecordsID = recordCount - 1;
--- a/app.ts
+++ b/app.ts
@@ -87,14 +87,8 @@ window.onload = function() {
 //// Targets heading div to creating the 1st row(Column names)
 
 function colHeading(heading: string) {
-    if (heading == "ID") {
-        let headingCol = `<div id="col_heading" class="col_heading">${heading}<input placeholder="Press Enter to search" id="idInput"></div>`;
-        headings.innerHTML += headingCol;
-    }
-    else {
-        let headingCol = `<div id="col_heading" class="col_heading">${heading}</div>`;
-        headings.innerHTML += headingCol;
-    }
+    let headingCol = `<div id="col_heading" class="col_heading">${heading}</div>`;
+    headings.innerHTML += headingCol;
 }
 
 //// Targets content div to create the actual table and fill with data
@@ -122,39 +116,39 @@ function clearTable() {
     content_cols.innerHTML = clear
 }
 
-// ID Search
+// // ID Search
 
-window.addEventListener("keydown", (e) => {
-    let key = e.key;
+// window.addEventListener("keydown", (e) => {
+//     let key = e.key;
 
-    if (key === "Enter") {
-        let search: any = document.querySelector('#idInput');
-        clearTable()
-        if (search.value != "" && search.value < 1000000 && search.value >= 0) {
-            pageStats.innerHTML = clear
-            let searchStats = "Showing result with ID of " + search.value;
-            pageStats.innerHTML = searchStats;
-            fetch("http://localhost:2050/records?from=" + search.value + "&to=" + search.value, {
-                method: "GET",
-                headers: { "Content-Type": "application/json" },
-            })
-            .then((res) => res.text())
-            .then((data) => {
-                data = JSON.parse(data);
-                let contentList = data;
-                for (let i = 0; i < contentList.length; i++) {
-                    contentNeeded.push(contentList[i])
-                    cols(contentList[i])
-                }
-            });
-        } else {
-            alert("There are no results with that ID!")
-            stats()
-            getTable()
-        }
-        search.value = clear
-    }
-})
+//     if (key === "Enter") {
+//         let search: any = document.querySelector('#idInput');
+//         clearTable()
+//         if (search.value != "" && search.value < 1000000 && search.value >= 0) {
+//             pageStats.innerHTML = clear
+//             let searchStats = "Showing result with ID of " + search.value;
+//             pageStats.innerHTML = searchStats;
+//             fetch("http://localhost:2050/records?from=" + search.value + "&to=" + search.value, {
+//                 method: "GET",
+//                 headers: { "Content-Type": "application/json" },
+//             })
+//             .then((res) => res.text())
+//             .then((data) => {
+//                 data = JSON.parse(data);
+//                 let contentList = data;
+//                 for (let i = 0; i < contentList.length; i++) {
+//                     contentNeeded.push(contentList[i])
+//                     cols(contentList[i])
+//                 }
+//             });
+//         } else {
+//             alert("There are no results with that ID!")
+//             stats()
+//             getTable()
+//         }
+//         search.value = clear
+//     }
+// })
 
 // Jump to ID  
 
@@ -273,8 +267,8 @@ nextButton.addEventListener("click", () => {
     clearTimeout(nextBtnStatus)
     nextCounter += 1
     nextBtnStatus = true
-    nextTurnFalse = setTimeout(nextBtnFalse, 2000)
-    nextActCheck = setTimeout(nextCheck, 2100)
+    nextTurnFalse = setTimeout(nextBtnFalse, 100)
+    nextActCheck = setTimeout(nextCheck, 200)
 })
 
 function nextBtnFalse() {
@@ -341,8 +335,8 @@ prevButton.addEventListener("click", () => {
     clearTimeout(prevBtnStatus)
     prevCounter += 1
     prevBtnStatus = true
-    prevTurnFalse = setTimeout(prevBtnFalse, 2000)
-    prevActCheck = setTimeout(prevCheck, 2100)
+    prevTurnFalse = setTimeout(prevBtnFalse, 100)
+    prevActCheck = setTimeout(prevCheck, 200)
 })
 
 function prevBtnFalse() {
--- a/app.ts
+++ b/app.ts
@@ -123,7 +123,6 @@ function colHeading(heading: string) {
 
 //// Targets content div to create the actual table and fill with data
 
-let columnRowData: any = []
 
 function cols(content: any) {
     let rows = `<div id=row-${content[0]} class="rows"></div>`;
@@ -381,13 +380,21 @@ function resizing() {
 }
 
 let runit: any;
+let resizeInProgress: any = false
+let resizeStatus: any;
 
 window.addEventListener("resize", () => {
     clearTimeout(runit);
-    runit = setTimeout(resizing, 100)
+    clearTimeout(resizeStatus)
+    resizeInProgress = true
+    resizeStatus = setTimeout(resizeStatusFalse, 100)
+    runit = setTimeout(resizing, 200)
     
 });
 
+function resizeStatusFalse() {
+    resizeInProgress = false
+}
 
 function prevRe() {
     let windowHeight = window.innerHeight;
--- a/app.ts
+++ b/app.ts
@@ -191,7 +191,9 @@ function idJump() {
     } 
     else if ( searchOne > 999990 && searchOne < 1000000 && searchOne >= 0 && searchOne !== 'e') {
         clearTable()
-        fetch("http://localhost:2050/records?from=" + searchOne + "&to=" + 999999, {
+        let lastParamOne = 999999 - toConvert
+
+        fetch("http://localhost:2050/records?from=" + lastParamOne + "&to=" + 999999, {
             method: "GET",
             headers: { "Content-Type": "application/json" },
         })
@@ -212,17 +214,12 @@ function idJump() {
         .then((data) => {
             data = JSON.parse(data);
             let count = data;
-            let currentStats = "Showing results from ID " + searchOne + " to " + 999999 + " out of " + count + " results.";
+            let currentStats = "Showing results from ID " + lastParamOne + " to " + 999999 + " out of " + count + " results.";
             pageStats.innerHTML = currentStats;
         })
     }
      else {
         alert("There are no records with that ID! Please check that your input does not exceed 999 999 and is not a negative amount.")
-        clearTable()
-        paramOne = 0
-        paramTwo = 9
-        stats()
-        getTable()
     }
     search.value = clear
 }
@@ -233,9 +230,7 @@ function idJump() {
 
 function next() {
     if (paramTwo == 999999) {
-        paramOne = "0"
-
-        resizing()
+        alert("You have reached the final page")
     } else {
         let nextAmount: any = (paramTwo - paramOne) + 1
 
@@ -277,8 +272,8 @@ function prev() {
 
     if (windowHeight < 600 && windowHeight >= 480) {
         if (paramOne <= 6) {
-            paramOne = 999993
-            prevRe()
+            paramOne = 0
+            alert("This is the first page")
         } else {
             let prevAmount: any = (paramTwo - paramOne) + 1
     
@@ -295,9 +290,8 @@ function prev() {
     } 
     else if (windowHeight < 480 && windowHeight >= 400) {
         if (paramOne <= 4) {
-            paramOne = 999995
-            
-            prevRe()
+            paramOne = 0
+            alert("This is the first page")
         } else {
             let prevAmount: any = (paramTwo - paramOne) + 1
     
@@ -312,9 +306,8 @@ function prev() {
     }
     else if (windowHeight < 400 && windowHeight > 300) {
         if (paramOne <= 2) {
-            paramOne = 999997
-            
-            prevRe()
+            paramOne = 0
+            alert("This is the first page")
         } else {
             let prevAmount: any = (paramTwo - paramOne) + 1
     
@@ -329,9 +322,8 @@ function prev() {
     }
     else if (windowHeight <= 300) {
         if (paramOne <= 1) {
-            paramOne = 999998
-            
-            prevRe()
+            paramOne = 0
+            alert("This is the first page")
         } else {
             let prevAmount: any = (paramTwo - paramOne) + 1
     
@@ -346,9 +338,8 @@ function prev() {
     }
     else {
         if (paramOne <= 9) {
-            paramOne = 999990
-            
-            prevRe()
+            paramOne = 0
+            alert("This is the first page")
         } else {
             let prevAmount: any = (paramTwo - paramOne) + 1
     
@@ -441,7 +432,7 @@ function resizing() {
     }
     else {
         if (paramOne > 999990) {
-            paramOne = 0
+            paramOne = 999990
         } else {
             //pass
         }
@@ -463,7 +454,6 @@ window.addEventListener("resize", () => {
     resizeInProgress = true
     resizeStatus = setTimeout(resizeStatusFalse, 100)
     runit = setTimeout(resizing, 200)
-    
 });
 
 function resizeStatusFalse() {
--- a/app.ts
+++ b/app.ts
@@ -4,6 +4,10 @@ let recordCount: number;
 
 //// On Window Load
 window.onload = function () {
+    const next = new Next(); // next class
+    const prev = new Prev(); // prev class
+    window.addEventListener("input", debounce(idJump, 500));
+    window.addEventListener("resize", debounce(resizing, 500));
     getRecordCount()
         .then(() => {
             return getHeadings();
@@ -11,10 +15,10 @@ window.onload = function () {
         .then(() => {
             return getTable();
         });
-    console.log("CALL HECTOR!!!");
 };
 
-function getParameters(fromParameter: number) {
+// Gets The 2nd paramater from the 1st parameter
+function getParameters(fromParameter: number): number {
     let toParameter: number;
     let noOfRows = getNoOfRows();
 
@@ -22,13 +26,14 @@ function getParameters(fromParameter: number) {
     return toParameter;
 }
 
+// Gets the number of rows on the screen
 function getNoOfRows(): number {
     const height = window.innerHeight;
     let noOfRows = Math.floor(height / 40);
     return noOfRows;
 }
 
-//// Functions To Create/Clear The HTML
+//// Functions To Create The HTML
 // Heading Row
 function createHeadingRow(headingData: string) {
     const heading: HTMLElement | null = document.getElementById("heading");
@@ -40,17 +45,19 @@ function createHeadingRow(headingData: string) {
 }
 
 // Table Content
-function createTableContent(contentData: string) {
+function createTableContent(contentData: any) {
     const content: HTMLElement | null = document.getElementById("content");
 
     let table = `<div id="row-${contentData[0]}" class="rows"></div>`;
     if (content !== null) {
         content.innerHTML += table;
+    } else {
+        alert("ERROR");
     }
 
     let rows: HTMLElement | null = document.getElementById("row-" + contentData[0]);
-    for (let x = 0; x < contentData.length; x++) {
-        let rowCols = `<div class="row_cols">${contentData[x]}</div>`;
+    for (let x of contentData) {
+        let rowCols = `<div class="row_cols">${x}</div>`;
         if (rows !== null) {
             rows.innerHTML += rowCols;
         }
@@ -67,19 +74,18 @@ function handleErrors(response: Response) {
 }
 
 // Heading Row (Getting the columns data)
-function getHeadings() {
+function getHeadings(): Promise<void> {
     return fetch("http://localhost:2050/columns", {
         method: "GET",
         headers: { "Content-Type": "application/json" },
     })
         .then(handleErrors)
-        .then((response) => response.text())
-        .then((data) => {
-            let headingData = JSON.parse(data);
-            for (let i = 0; i < headingData.length; i++) {
-                createHeadingRow(headingData[i]);
+        .then((response: Response) => response.json())
+        .then((data: string) => {
+            let headingData = data;
+            for (let heading of headingData) {
+                createHeadingRow(heading);
             }
-            getParameters(0);
         })
         .catch((error: Error) => {
             console.log(error);
@@ -87,7 +93,7 @@ function getHeadings() {
 }
 
 // Table Content (Getting the table's data)
-function getTable() {
+function getTable(): Promise<void> {
     const content: HTMLElement | null = document.getElementById("content");
     let toParameter = getParameters(fromParameter);
     const pageStats: HTMLElement | null = document.getElementById("pageStats");
@@ -108,11 +114,11 @@ function getTable() {
         headers: { "Content-Type": "application/json" },
     })
         .then(handleErrors)
-        .then((res: Response) => res.text())
-        .then((data) => {
-            let contentData = JSON.parse(data);
-            for (let i = 0; i < contentData.length; i++) {
-                createTableContent(contentData[i]);
+        .then((res: Response) => res.json())
+        .then((data: string) => {
+            let contentData = data;
+            for (let content of contentData) {
+                createTableContent(content);
             }
         })
         .catch((error: Error) => {
@@ -121,15 +127,15 @@ function getTable() {
 }
 
 // Gets Total Of All Records
-function getRecordCount() {
+function getRecordCount(): Promise<void> {
     return fetch("http://localhost:2050/recordCount", {
         method: "GET",
         headers: { "Content-Type": "application/json" },
     })
         .then(handleErrors)
-        .then((res: Response) => res.text())
-        .then((data) => {
-            recordCount = JSON.parse(data);
+        .then((res: Response) => res.json())
+        .then((data: number) => {
+            recordCount = data;
         })
         .catch((error: Error) => {
             console.log(error);
@@ -155,131 +161,128 @@ function resizing() {
     if (end > maxRecordsID) {
         toParameter = maxRecordsID;
         fromParameter = toParameter - getNoOfRows();
-    } else {
-        toParameter = fromParameter + getNoOfRows();
     }
     getTable();
 }
 
-window.addEventListener("resize", debounce(resizing, 500));
-
 //// Navigation
 // Next
-{
-    let nextCount = 0;
-    const nextButton: HTMLElement | null = document.getElementById("next");
+class Next {
+    nextCount: number;
+    nextButton: HTMLElement | null;
 
-    const nextDebounce = (fn: any, delay: number) => {
-        return function () {
-            nextCount++;
-            clearTimeout(timer);
-            timer = setTimeout(() => {
-                fn();
-            }, delay);
+    constructor() {
+        this.nextCount = 0;
+        this.nextButton = document.getElementById("next");
+
+        const nextDebounce = (fn: any, delay: number) => {
+            return () => {
+                this.nextCount++;
+                clearTimeout(timer);
+                timer = setTimeout(() => {
+                    fn();
+                }, delay);
+            };
         };
-    };
 
-    let next = () => {
-        let toParameter = getParameters(fromParameter);
-        let maxRecordsID = recordCount - 1;
+        let next = () => {
+            let toParameter = getParameters(fromParameter);
+            let maxRecordsID = recordCount - 1;
 
-        if (toParameter === maxRecordsID) {
-            alert("You have reached the final page");
-        }
+            if (toParameter === maxRecordsID) {
+                alert("You have reached the final page");
+            }
 
-        let nextAmount = toParameter - fromParameter + 1;
-        let nextCountAmount = nextAmount * nextCount;
-        fromParameter = fromParameter + nextCountAmount;
-        toParameter = fromParameter + getNoOfRows();
+            let nextAmount = toParameter - fromParameter + 1;
+            let nextCountAmount = nextAmount * this.nextCount;
+            fromParameter = fromParameter + nextCountAmount;
+            toParameter = fromParameter + getNoOfRows();
 
-        let end = fromParameter + getNoOfRows();
+            let end = fromParameter + getNoOfRows();
 
-        if (end > maxRecordsID) {
-            toParameter = maxRecordsID;
-            fromParameter = toParameter - getNoOfRows();
-        }
+            if (end > maxRecordsID) {
+                toParameter = maxRecordsID;
+                fromParameter = toParameter - getNoOfRows();
+            }
 
-        nextCount = 0;
+            this.nextCount = 0;
 
-        getTable();
-    };
+            getTable();
+        };
 
-    if (nextButton !== null) {
-        nextButton.addEventListener("click", nextDebounce(next, 500));
+        this.nextButton?.addEventListener("click", nextDebounce(next, 500));
     }
 }
 
 // Previous
-{
-    let prevCount = 0;
-    const prevButton: HTMLElement | null = document.getElementById("prev");
+class Prev {
+    prevCount: number;
+    prevButton: HTMLElement | null;
 
-    const prevDebounce = (fn: any, delay: number) => {
-        return function () {
-            prevCount++;
-            clearTimeout(timer);
-            timer = setTimeout(() => {
-                fn();
-            }, delay);
+    constructor() {
+        this.prevCount = 0;
+        this.prevButton = document.getElementById("prev");
+
+        const prevDebounce = (fn: any, delay: number) => {
+            return () => {
+                this.prevCount++;
+                clearTimeout(timer);
+                timer = setTimeout(() => {
+                    fn();
+                }, delay);
+            };
         };
-    };
 
-    let prev = () => {
-        let toParameter = getParameters(fromParameter);
+        let prev = () => {
+            let toParameter = getParameters(fromParameter);
 
-        if (fromParameter === 0) {
-            alert("You Are On The First Page");
-        } else {
-            let prevAmount = toParameter - fromParameter + 1;
-            let prevCountAmount = prevAmount * prevCount;
-
-            let intOne = fromParameter - prevCountAmount;
-
-            if (intOne < 0) {
-                fromParameter = 0;
+            if (fromParameter === 0) {
+                alert("You Are On The First Page");
             } else {
-                fromParameter = intOne;
+                let prevAmount = toParameter - fromParameter + 1;
+                let prevCountAmount = prevAmount * this.prevCount;
+
+                let intOne = fromParameter - prevCountAmount;
+
+                if (intOne < 0) {
+                    fromParameter = 0;
+                } else {
+                    fromParameter = intOne;
+                }
+
+                this.prevCount = 0;
+
+                getTable();
             }
+        };
 
-            prevCount = 0;
-
-            getTable();
-        }
-    };
-
-    if (prevButton !== null) {
-        prevButton.addEventListener("click", prevDebounce(prev, 500));
+        this.prevButton?.addEventListener("click", prevDebounce(prev, 500));
     }
 }
 
 // ID Jump
-{
+function idJump() {
     const input: any = document.querySelector("input");
+    let toParameter = getParameters(fromParameter);
+    let currentID = fromParameter;
+    let search = input.value;
+    let end = parseInt(search) + getNoOfRows();
+    let maxRecordsID = recordCount - 1;
 
-    let idJump = () => {
-        let toParameter = getParameters(fromParameter);
-        let currentID = fromParameter;
-        let search = input.value;
-        let end = parseInt(search) + getNoOfRows();
-        let maxRecordsID = recordCount - 1;
-
-        if (search !== NaN && search !== "" && search <= maxRecordsID && search >= 0) {
-            if (end > maxRecordsID) {
-                toParameter = maxRecordsID;
-                fromParameter = toParameter - getNoOfRows();
-            } else {
-                fromParameter = parseInt(search);
-                toParameter = fromParameter + getNoOfRows();
-            }
-        } else if (search !== "") {
-            alert("Make Sure Your Desired ID Is Not A Negative Number Or Doesn't Exceed 999999");
-            fromParameter = currentID;
+    if (search !== NaN && search !== "" && search <= maxRecordsID && search >= 0) {
+        if (end > maxRecordsID) {
+            toParameter = maxRecordsID;
+            fromParameter = toParameter - getNoOfRows();
+        } else {
+            fromParameter = parseInt(search);
             toParameter = fromParameter + getNoOfRows();
-            input.value = "";
         }
+    } else if (search !== "") {
+        alert("Make Sure Your Desired ID Is Not A Negative Number Or Doesn't Exceed 999999");
+        fromParameter = currentID;
+        toParameter = fromParameter + getNoOfRows();
+        input.value = "";
+    }
 
-        getTable();
-    };
-
-    window.addEventListener("input", debounce(idJump, 500));
+    getTable();
 }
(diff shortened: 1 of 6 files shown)


app.ts
```diff
@@ -0,0 +1,473 @@
+const headings: any = document.querySelector("#Headings");
+const content_cols: any = document.querySelector("#Content");
+const pageStats: any = document.getElementById('pageStats')
+const nextButton: any = document.querySelector("#next");
+const prevButton: any = document.querySelector("#prev");
+const clear = "";
+let paramOne: any = "0"
+let paramTwo: any = "9"
```
geraldtivatyi: I think rather calculate this value based on the size of the window than have a fixed value. Screen sizes change and you can potentially fit more records on the screen. Try to fit as much readable records on the screen as possible.

anzelIMQS (changes requested):
Before I actually review the code, please add semicolons at the end of each line. You have a lot of them missing.

DevinFledermaus:
all done

app.ts
```diff
@@ -0,0 +1,462 @@
+const headings: any = document.querySelector("#Headings");
+const content_cols: any = document.querySelector("#Content");
+const pageStats: any = document.getElementById("pageStats");
+const nextButton: any = document.querySelector("#next");
+const prevButton: any = document.querySelector("#prev");
+const clear = "";
+let paramOne: any = "0";
+let paramTwo: any = "9";
```
anzelIMQS: Give your variables a more definitive type. Refrain from using any if not necessary. Why not make them **numbers** type if you are doing calculations on them. No need to parse back-and-forth everytime. Just once when you actually need them to be strings.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+let contentNeeded: any = [];
+
+// Fetch requests
+
+//// Heading Row(The Columns)
+
+function getColumns() {
```
anzelIMQS: Add a return type. In this case void. The function keyword is also not entirely necessary.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+  fetch("http://localhost:2050/columns", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((res) => res.text())
+    .then((data) => {
```
anzelIMQS: Implement a catch to catch an error. i.e. At least have logging of what went wrong.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+
+//// Table Content
+
+function getTable() {
+  fetch("http://localhost:2050/records?from=" + paramOne + "&to=" + paramTwo, {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((res) => res.text())
+    .then((data) => {
```
anzelIMQS: Add error catching

app.ts
```diff
@@ -0,0 +1,462 @@
...
+        " out of " +
+        count +
+        " results.";
+      pageStats.innerHTML = currentStats;
+    });
+}
+
+// Loads all intial fields/data
+
+window.onload = function () {
```
anzelIMQS: Use an arrow function.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+  let windowHeight = window.innerHeight;
+
+  if (windowHeight < 600 && windowHeight >= 480) {
+    paramOne = paramOne;
```
anzelIMQS: Seems unnecessary to set one to itself?

app.ts
```diff
@@ -0,0 +1,462 @@
+const headings: any = document.querySelector("#Headings");
+const content_cols: any = document.querySelector("#Content");
+const pageStats: any = document.getElementById("pageStats");
+const nextButton: any = document.querySelector("#next");
+const prevButton: any = document.querySelector("#prev");
+const clear = "";
+let paramOne: any = "0";
+let paramTwo: any = "9";
+let contentNeeded: any = [];
+
```
geraldtivatyi: There are too many global variables here which is generally frowned upon. There are few variables that are acceptable as global variables.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+    getColumns();
+    getTable();
+  } else if (windowHeight <= 300) {
+    paramOne = paramOne;
+    paramTwo = parseInt(paramOne) + 1;
+    paramTwo = paramTwo.toString();
+    stats();
+    getColumns();
+    getTable();
+  } else {
```
anzelIMQS: You can now also remove the last **if** if you run those 3 lines afterward.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+
+  if (windowHeight < 600 && windowHeight >= 480) {
+    paramOne = paramOne;
+    paramTwo = parseInt(paramOne) + 6;
+    paramTwo = paramTwo.toString();
+    stats();
+    getColumns();
+    getTable();
+  } else if (windowHeight < 480 && windowHeight >= 400) {
+    paramOne = paramOne;
```
anzelIMQS: Same as above case.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+window.onload = function () {
+  let windowHeight = window.innerHeight;
+
+  if (windowHeight < 600 && windowHeight >= 480) {
+    paramOne = paramOne;
+    paramTwo = parseInt(paramOne) + 6;
+    paramTwo = paramTwo.toString();
+    stats();
+    getColumns();
+    getTable();
```
anzelIMQS: These lines are repetitive in all your cases and require no logic. This means you can take them out and run them after your if statements.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+  } else if (windowHeight < 480 && windowHeight >= 400) {
+    paramOne = paramOne;
+    paramTwo = parseInt(paramOne) + 4;
+    paramTwo = paramTwo.toString();
+    stats();
+    getColumns();
+    getTable();
```
anzelIMQS: Same as above case.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+  } else if (windowHeight < 400 && windowHeight > 300) {
+    paramOne = paramOne;
```
anzelIMQS: Same as above case.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+    paramTwo = parseInt(paramOne) + 2;
+    paramTwo = paramTwo.toString();
+    stats();
+    getColumns();
+    getTable();
```
anzelIMQS: Same as above case.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+  } else if (windowHeight <= 300) {
+    paramOne = paramOne;
```
anzelIMQS: Same as above case.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+    paramTwo = parseInt(paramOne) + 1;
+    paramTwo = paramTwo.toString();
+    stats();
+    getColumns();
+    getTable();
```
anzelIMQS: Same as above case.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+      for (let i = 0; i < contentList.length; i++) {
+        let contentListContent: any = contentList[i];
+        cols(contentListContent);
+      }
+    });
+}
+
+//// Display the current shown results
+
+function stats() {
```
anzelIMQS: Add a return type. In this case void. The function keyword is also not entirely necessary.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+    getColumns();
+    getTable();
+  }
+};
+
+// Displaying data into html
+
+//// Targets heading div to creating the 1st row(Column names)
+
+function colHeading(heading: string) {
```
anzelIMQS: Add a return type. In this case void. The function keyword is also not entirely necessary.

server/main.go
```diff
@@ -14,10 +14,10 @@ import (
 )
 
 const recordCount = 1000000
-const columnCount = 11
+const columnCount = 12
 const delayResponse = 500 * time.Millisecond
 
-var columns = [columnCount]string{"ID", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J"}
+var columns = [columnCount]string{"ID", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K"}
```
anzelIMQS: I don't think it should be necessary to touch the backend. Only frontend.

anzelIMQS (changes requested):

app.ts
```diff
@@ -0,0 +1,462 @@
...
+      pageStats.innerHTML = currentStats;
+    });
+}
+
+// Loads all intial fields/data
+
+window.onload = function () {
+  let windowHeight = window.innerHeight;
+
+  if (windowHeight < 600 && windowHeight >= 480) {
```
geraldtivatyi: Hmmm, I would avoid doing this check: again it ties into the calculation to fit as much readable rows on the screen. So rather have a function that calculates and returns the number of rows you can fit on the page and then use that value to set your parameters.

app.ts
```diff
@@ -0,0 +1,462 @@
+const headings: any = document.querySelector("#Headings");
+const content_cols: any = document.querySelector("#Content");
+const pageStats: any = document.getElementById("pageStats");
+const nextButton: any = document.querySelector("#next");
+const prevButton: any = document.querySelector("#prev");
+const clear = "";
+let paramOne: any = "0";
```
geraldtivatyi: Use the `any` type sparingly. Check the javascript style guide for the preferences and standards that IMQS uses.

app.ts
geraldtivatyi: Preferably use arrow functions

app.ts
geraldtivatyi: As Anzel said, these lines are repeated many times. Its a good sign to create a new method or just move code outside of if-esle statements if you see yourselfrepeating it without any unique logic to it.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+}
+
+// Loads all intial fields/data
+
+window.onload = function () {
+  let windowHeight = window.innerHeight;
+
+  if (windowHeight < 600 && windowHeight >= 480) {
+    paramOne = paramOne;
+    paramTwo = parseInt(paramOne) + 6;
```
geraldtivatyi: Check Javascript Style Guide for how to do type casting.

app.ts
```diff
@@ -0,0 +1,462 @@
...
+  let convert = parseInt(searchOne) + toConvert;
+  let searchTwo = convert.toString();
+  paramOne = searchOne;
+  paramTwo = searchTwo;
+
+  if (
+    searchOne !== "" &&
+    searchOne <= 999990 &&
+    searchOne >= 0 &&
+    searchOne !== "e"
```
geraldtivatyi: I think the standard way in IMQS is to put the if statement's condition on the same line. Follow up with Fritz on this.

geraldtivatyi (commented):
In general, go through the JavaScript Style Guide described in the project README. Also I don't see any error checking especially when you are making calls to a server but also when you call functions/methods.

CelesteNaude (changes requested):
Just a few changes from my side.

app.ts
CelesteNaude: As Gerald said, global variables are bad. You can move them to their own class.

index.html
```diff
@@ -3,10 +3,22 @@
...
+		<div id="jumpID"><input type="number" id="idJump" placeholder="Input a starting ID"><button onclick="idJump()">Jump To ID</button></div>
+		<div class="btns">
+			<button class="prev" id="prev">Prev</button>
+			<button class="next" id="next">Next</button>
+		</div>
+	</div>
+	<div id="Headings"></div>
+	<div id="Content"></div>
+
+	<script src="app.js"></script>
```
CelesteNaude: Your `script` tag should be in the `head` tag.

index.html
```diff
@@ -3,10 +3,22 @@
...
 	<title>JS Onboard Project</title>
 	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+	<link rel="stylesheet" href="style.css">
 </head>
 
 <body>
-	<p>Hello</p>
+	<div class="nav">
+		<label id="pageStats"></label>
+		<div id="jumpID"><input type="number" id="idJump" placeholder="Input a starting ID"><button onclick="idJump()">Jump To ID</button></div>
```
CelesteNaude: You should rather specify your `onclick` events in your TypeScript code.

CelesteNaude (changes requested):

app.ts
```diff
@@ -0,0 +1,277 @@
...
+      headers: { "Content-Type": "application/json" },
+    })
+      .then((res) => res.text())
+      .then((data) => {
+        data = JSON.parse(data);
+        let count = data;
+        let currentStats = "Showing results from " + paramOne + " to " + paramTwo + " out of " + count + " results.";
+        pageStats.innerHTML = currentStats;
+      });
+  } catch (error) {}
```
anzelIMQS: Same as the above comment on the catch

app.ts
```diff
@@ -0,0 +1,277 @@
...
+      .then((data) => {
+        data = JSON.parse(data);
+        let headingData = data;
+        for (let i = 0; i < headingData.length; i++) {
+          createHeadingRow(headingData[i]);
+        }
+      });
+  } catch (error) {
+    console.log(error);
+  }
```
anzelIMQS: I meant, a catch on the **then** of the Promise.
e.g. `then( () => {...}).catch( (error) => {})`

Not a seperate try and catch block.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+      .then((data) => {
+        data = JSON.parse(data);
+        let contentData = data;
+        for (let i = 0; i < contentData.length; i++) {
+          createTableContent(contentData[i]);
+        }
+      });
+  } catch (error) {
+    console.log(error);
+  }
```
anzelIMQS: Same as the above comment on the catch

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  try {
+    fetch("http://localhost:2050/recordCount", {
+      method: "GET",
+      headers: { "Content-Type": "application/json" },
+    })
+      .then((res) => res.text())
+      .then((data) => {
+        data = JSON.parse(data);
+        let count = data;
+        let currentStats = "Showing results from " + paramOne + " to " + paramTwo + " out of " + count + " results.";
```
anzelIMQS: Rather use Template Literals. ${}

app.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  try {
+    fetch("http://localhost:2050/recordCount", {
+      method: "GET",
+      headers: { "Content-Type": "application/json" },
+    })
+      .then((res) => res.text())
+      .then((data) => {
+        data = JSON.parse(data);
+        let count = data;
```
anzelIMQS: Probably not necessary to do this over 2 lines.

let count = JSON.parse(data);

app.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (end > 999999) {
+      paramTwo = 999999;
+      paramOne = paramTwo - getNoOfRows();
+    } else {
+      paramOne = parseInt(search);
+      paramTwo = paramOne + getNoOfRows();
+    }
+  } else if (search === "") {
+    //pass
+  } else {
```
anzelIMQS: Try to never implement an if-else statement that does nothing. Remove the last else and then you can change your logic for the last if-else like **(search !== "")** because that is in essence what the logic is for the last else in your case.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  } catch (error) {
+    console.log(error);
+  }
+};
+
+// Table Content (Getting the table's data)
+
+const getTable = () => {
+  try {
+    fetch("http://localhost:2050/records?from=" + paramOne + "&to=" + paramTwo, {
```
anzelIMQS: I'd like you to change this function to accept the from and to variables as parameters. We don't like global variables if not necessary. 

Also give these variables more descriptive names. Give paramOne and paramTwo names more meaning to what they really are. 
A tip: from and to 😉

app.ts
```diff
@@ -0,0 +1,277 @@
...
+        }
+      });
+  } catch (error) {
+    console.log(error);
+  }
+};
+
+// Displays The Current Results Being Shown
+
+const stats = () => {
```
anzelIMQS: In general, this function should not have to do a recordCount API call. The recordCount is constant throughout in this case. You can probably just build your string in here and update it.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  nextCount = 0;
+
+  clearTable();
+  getTable();
+  stats();
+};
+
+next = nextDebounce(next, 500);
```
anzelIMQS: You should not have to do this. Just pass the nextDebounce in directly to the eventListener in this case.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  getHeadings();
+  getTable();
+  stats();
+};
+
+//// Navigation
+
+// Next
+const nextButton: any = document.querySelector("#next");
+let nextCount = 0;
```
anzelIMQS: I'm struggling to understand why you need these count variables? They look like they are kinda contradicting the point of the debouncing?

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  stats();
+};
+
+next = nextDebounce(next, 500);
+
+nextButton.addEventListener("click", next);
+
+// Previous
+const prevButton: any = document.querySelector("#prev");
+let prevCount = 0;
```
anzelIMQS: Same as the above comment for the next

app.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    prevCount = 0;
+
+    clearTable();
+    getTable();
+    stats();
+  }
+};
+
+prev = prevDebounce(prev, 500);
```
anzelIMQS: Same vibe as the above next

app.ts
```diff
@@ -0,0 +1,277 @@
...
+    paramTwo = paramOne + getNoOfRows();
+    input.value = "";
+  }
+
+  clearTable();
+  stats();
+  getTable();
+};
+
+idJump = debounce(idJump, 500);
```
anzelIMQS: Same vibe as the above next.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  return function () {
+    nextCount++;
+    clearTimeout(timer);
+    timer = setTimeout(() => {
+      fn();
+    }, delay);
+  };
+};
+
+let next = () => {
```
anzelIMQS: The paramOne and paramTwo should be accepted as parameters to this function.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+  return function () {
+    prevCount++;
+    clearTimeout(timer);
+    timer = setTimeout(() => {
+      fn();
+    }, delay);
+  };
+};
+
+let prev = () => {
```
anzelIMQS: Same as mentioned in the next.

app.ts
```diff
@@ -0,0 +1,277 @@
...
+//// Sizing And Resizing
+
+let resizing = () => {
+  let end = paramOne + getNoOfRows();
+
+  if (end > 999999) {
+    paramTwo = 999999;
+    paramOne = paramTwo - getNoOfRows();
+  } else {
+    paramOne;
```
anzelIMQS: Remove redundancy.

anzelIMQS (changes requested):
Function restructuring to remove the global parameters. These values should be passed around. Functions should not have to know about stuff happening outside of their scope.

app.ts
DevinFledermaus: I'm using it to get the amount of times the button is being clicked so i can move next according to the amount

app.ts
```diff
@@ -0,0 +1,276 @@
+let fromParameter = 0;
+
+const getParameters = (fromParameter: number) => {
+  let toParameter: number;
+  const height = window.innerHeight;
+
+  let noOfRows = Math.floor(height / 40);
```
geraldtivatyi: What happens if you rather call the method `getNoOfRows` which you created, instead of repeating the calculation here. This will also mean you won't need the variable `height`.

app.ts
```diff
@@ -0,0 +1,272 @@
...
+        toParameter = fromParameter + getNoOfRows();
+      }
+    } else if (search !== "") {
+      alert("Make Sure Your Desired ID Is Not A Negative Number Or Doesn't Exceed 999999");
+      fromParameter = currentID;
+      toParameter = fromParameter + getNoOfRows();
+      input.value = "";
+    } else {
+      //pass
+    }
```
anzelIMQS: I mean. You can completely remove this else since is does nothing.

app.ts
anzelIMQS: What I mean by recordCount is constant is that you don't have to make an API call every time you call **stats** since the **value** is going to be the same for each call. 

In this case, I recommend you create a global const variable that you assign the recordCount API response once on the initial load of your program. At the moment it is a hardcoded value. Let the backend determine it.

app.ts
```diff
@@ -0,0 +1,272 @@
...
+    .then((res) => res.text())
+    .then((data) => {
+      let headingData = JSON.parse(data);
+      for (let i = 0; i < headingData.length; i++) {
+        createHeadingRow(headingData[i]);
+      }
+      getParameters((fromParameter = 0));
+    })
+    .catch((error) => {
+      console.log("Call Hector", error);
```
anzelIMQS: Who's Hector?

app.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  if (end > 999999) {
+    toParameter = 999999;
+    fromParameter = toParameter - getNoOfRows();
+  } else {
+    toParameter = fromParameter + getNoOfRows();
+  }
+  clearTable();
+  getTable();
+  stats();
```
anzelIMQS: I can definitely see these 3 functions actually being in one as it always goes together logically. They are always repeated in that order and are not too complex to be split up. It can be all together in getTable()?

app.ts
```diff
@@ -0,0 +1,272 @@
...
+  let toParameter = getParameters(fromParameter);
+
+  let currentStats = `Showing results from ${fromParameter} to ${toParameter} out of 1 000 000 results.`;
+  pageStats.innerHTML = currentStats;
+};
+
+//// Debounce
+
+const debounce = (fn: any, delay: number) => {
+  let timer: number;
```
anzelIMQS: You might need to add timer as a global variable. It's something I noticed with Ashton as well and he got it to work. At this stage, timer is only in the scope of this function.

anzelIMQS (changes requested):
A few more changes regarding functions.

app.ts
```diff
@@ -0,0 +1,274 @@
...
+  };
+};
+
+//// Sizing And Resizing
+
+let resizing = () => {
+  let end = fromParameter + getNoOfRows();
+  let toParameter: number;
+
+  if (end > 999999) {
```
anzelIMQS: Don't use 999999. You have recordCount for the max..

app.ts
```diff
@@ -0,0 +1,274 @@
...
+      recordCount = JSON.parse(data);
+    })
+    .catch((error) => {
+      console.log(error);
+    });
+};
+
+// getRecordCount();
+// getHeadings();
+// getTable();
```
anzelIMQS: Remove clutter.

app.ts
```diff
@@ -0,0 +1,274 @@
...
+      timer = setTimeout(() => {
+        fn();
+      }, delay);
+    };
+  };
+
+  let next = () => {
+    let toParameter = getParameters(fromParameter);
+
+    if (toParameter === 999999) {
```
anzelIMQS: Same vibe for recordCount

app.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    let nextAmount = toParameter - fromParameter + 1;
+    let nextCountAmount = nextAmount * nextCount;
+    fromParameter = fromParameter + nextCountAmount;
+    toParameter = fromParameter + getNoOfRows();
+
+    let end = fromParameter + getNoOfRows();
+
+    if (end > 999999) {
+      toParameter = 999999;
```
anzelIMQS: Same vibe for recordCount

app.ts
```diff
@@ -0,0 +1,274 @@
...
+    .then(() => {
+      return getHeadings();
+    })
+    .then(() => {
+      return getTable();
+    });
+  console.log("CALL HECTOR!!!");
+};
+
+const getParameters = (fromParameter: number) => {
```
anzelIMQS: Add return type

app.ts
```diff
@@ -0,0 +1,274 @@
...
+  let toParameter: number;
+  let noOfRows = getNoOfRows();
+
+  toParameter = fromParameter + noOfRows;
+  return toParameter;
+};
+
+const getNoOfRows = () => {
```
anzelIMQS: Add return type

app.ts
```diff
@@ -0,0 +1,274 @@
...
+
+      let intOne = fromParameter - prevCountAmount;
+
+      if (intOne < 0) {
+        fromParameter = 0;
+      } else {
+        fromParameter = intOne;
+      }
+
+      toParameter = fromParameter + getNoOfRows();
```
anzelIMQS: Is this still necessary? toParameter isn't used after this?

app.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  let idJump = () => {
+    let toParameter = getParameters(fromParameter);
+    let currentID = fromParameter;
+    let search = input.value;
+    let end = parseInt(search) + getNoOfRows();
+
+    if (search !== NaN && search !== "" && search < 1000000 && search >= 0) {
+      if (end > 999999) {
+        toParameter = 999999;
```
anzelIMQS: Same vibe for recordCount.

app.ts
```diff
@@ -0,0 +1,274 @@
...
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((res) => res.text())
+    .then((data) => {
+      let headingData = JSON.parse(data);
+      for (let i = 0; i < headingData.length; i++) {
+        createHeadingRow(headingData[i]);
+      }
+      getParameters((fromParameter = 0));
```
anzelIMQS: Was this necessary? getParameters returns your toParameter but you do nothing with the return result since you don't need it?

anzelIMQS (changes requested):
More cleaning up, please.

app.ts
```diff
@@ -0,0 +1,271 @@
...
+  const height = window.innerHeight;
+  let noOfRows = Math.floor(height / 40);
+  return noOfRows;
+};
+
+//// Functions To Create/Clear The HTML
+
+// Heading Row
+
+const createHeadingRow = (headingData: string) => {
```
FritzOnFire: Rather remove the empty line between the variable and the comment.

app.ts
```diff
@@ -0,0 +1,271 @@
...
+  const heading: any = document.querySelector("#heading");
```
FritzOnFire: The only place you can have an `any` in the project is when handling data that comes from the back-end. And even then you can actually use the real type instead of `any`.

app.ts
```diff
@@ -0,0 +1,271 @@
...
+    rows.innerHTML += rowCols;
+  }
+};
+
+//// Fetch Requests
+
+// Heading Row (Getting the columns data)
+
+const getHeadings = () => {
+  fetch("http://localhost:2050/columns", {
```
FritzOnFire: You should put a `return` in front of the `fetch`, that would allow other functions to wait on this function.

app.ts
```diff
@@ -0,0 +1,271 @@
...
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((res) => res.text())
+    .then((data) => {
+      let headingData = JSON.parse(data);
+      for (let i = 0; i < headingData.length; i++) {
+        createHeadingRow(headingData[i]);
+      }
+      getParameters((fromParameter = 0));
```
FritzOnFire: To many braces and I would rather have this as two lines, where you explicitly call `getParamter` with a hard coded `0`

app.ts
```diff
@@ -0,0 +1,271 @@
...
+//// Fetch Requests
+
+// Heading Row (Getting the columns data)
+
+const getHeadings = () => {
+  fetch("http://localhost:2050/columns", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((res) => res.text())
```
FritzOnFire: You should check for an error in `res` here. And you either want to remove the `()` around the `res` or specify the type, because currently the compiler thinks the type is `any`

app.ts
```diff
@@ -0,0 +1,271 @@
...
+    .then((data) => {
+      let headingData = JSON.parse(data);
+      for (let i = 0; i < headingData.length; i++) {
+        createHeadingRow(headingData[i]);
+      }
+      getParameters((fromParameter = 0));
+    })
+    .catch((error) => {
```
FritzOnFire: You want to remove the `()` around the `error` or specify the type, as currently the compiler thinks the type is `any`

FritzOnFire (changes requested):
Please convert all spaces to tabs. Please change the functions that you defined as variables to be defined as normal functions, you can leave the ones that you actually override as they were.

I made a comment about removing the empty line between the comment and the variable, there are a few places that this should happen.

Its a bit hard to review as it is, please make the changes where ever you think they are relevant and then I will review again.

FritzOnFire (changes requested):

app.ts
FritzOnFire: bump

app.ts
```diff
@@ -0,0 +1,285 @@
...
+//// On Window Load
+window.onload = function () {
+    getRecordCount()
+        .then(() => {
+            return getHeadings();
+        })
+        .then(() => {
+            return getTable();
+        });
+    console.log("CALL HECTOR!!!");
```
FritzOnFire: Stray console log.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+};
+
+function getParameters(fromParameter: number) {
```
FritzOnFire: Please add a return type

app.ts
```diff
@@ -0,0 +1,285 @@
...
+    const heading: HTMLElement | null = document.getElementById("heading");
+
+    let headings = `<div class="headings" id="headings">${headingData}</div>`;
+    if (heading !== null) {
+        heading.innerHTML += headings;
+    }
+}
+
+// Table Content
+function createTableContent(contentData: string) {
```
FritzOnFire: They way you are using `contentData` gives me the idea that it is not simply a `string` -_-

app.ts
```diff
@@ -0,0 +1,285 @@
...
+    const content: HTMLElement | null = document.getElementById("content");
+
+    let table = `<div id="row-${contentData[0]}" class="rows"></div>`;
+    if (content !== null) {
```
FritzOnFire: You probably want to stop the function at this point, and tell the user something is wrong, as the rest of this function is dependent on this having a value.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+
+// Heading Row (Getting the columns data)
+function getHeadings() {
+    return fetch("http://localhost:2050/columns", {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(handleErrors)
+        .then((response) => response.text())
+        .then((data) => {
```
FritzOnFire: Please remove the `()` or specify the type as currently the compiler thinks the type is `any`.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// Heading Row (Getting the columns data)
+function getHeadings() {
+    return fetch("http://localhost:2050/columns", {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(handleErrors)
+        .then((response) => response.text())
```
FritzOnFire: Rather use `.json()` to avoid parsing the json in the next section.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+        .then((data) => {
+            let headingData = JSON.parse(data);
+            for (let i = 0; i < headingData.length; i++) {
```
FritzOnFire: Change this to a `for-of` loop.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+// Response Error Handling
+function handleErrors(response: Response) {
+    if (!response.ok) {
+        throw Error(response.statusText);
+    }
+    return response;
+}
+
+// Heading Row (Getting the columns data)
+function getHeadings() {
```
FritzOnFire: Please add a return type.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+            }
+            getParameters(0);
+        })
+        .catch((error: Error) => {
+            console.log(error);
+        });
+}
+
+// Table Content (Getting the table's data)
+function getTable() {
```
FritzOnFire: Please add a return type.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(handleErrors)
+        .then((response) => response.text())
+        .then((data) => {
+            let headingData = JSON.parse(data);
+            for (let i = 0; i < headingData.length; i++) {
+                createHeadingRow(headingData[i]);
+            }
+            getParameters(0);
```
FritzOnFire: what is the purpose of calling this function here?

app.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (pageStats !== null) {
+        pageStats.innerHTML = currentStats;
+    }
+
+    return fetch(`http://localhost:2050/records?from=${fromParameter}&to=${toParameter}`, {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(handleErrors)
+        .then((res: Response) => res.text())
```
FritzOnFire: Rather use `.json()` to avoid parsing the json in the next section.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+        .then((data) => {
+            let contentData = JSON.parse(data);
+            for (let i = 0; i < contentData.length; i++) {
```
FritzOnFire: Change this to a `for-of` loop

app.ts
```diff
@@ -0,0 +1,285 @@
...
+        pageStats.innerHTML = currentStats;
+    }
+
+    return fetch(`http://localhost:2050/records?from=${fromParameter}&to=${toParameter}`, {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(handleErrors)
+        .then((res: Response) => res.text())
+        .then((data) => {
```
FritzOnFire: Please remove the `()` of specify the type, as the compiler currently thinks the type is `any`

app.ts
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// Gets Total Of All Records
+function getRecordCount() {
+    return fetch("http://localhost:2050/recordCount", {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(handleErrors)
+        .then((res: Response) => res.text())
```
FritzOnFire: Rather use `.json()` to avoid parsing the json in the next section.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+        .then((data) => {
```
FritzOnFire: Please remove the `()` or specify the type, as the compiler currently thinks the type is `any`

app.ts
```diff
@@ -0,0 +1,285 @@
...
+function resizing() {
+    let end = fromParameter + getNoOfRows();
+    let toParameter: number;
+    let maxRecordsID = recordCount - 1;
+
+    if (end > maxRecordsID) {
+        toParameter = maxRecordsID;
+        fromParameter = toParameter - getNoOfRows();
+    } else {
+        toParameter = fromParameter + getNoOfRows();
```
FritzOnFire: What is the purpose of this line?

app.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+    getTable();
+}
+
+window.addEventListener("resize", debounce(resizing, 500));
```
FritzOnFire: This line should not be in global space.

app.ts
```diff
@@ -0,0 +1,285 @@
...
+            fromParameter = currentID;
+            toParameter = fromParameter + getNoOfRows();
+            input.value = "";
+        }
+
+        getTable();
+    };
+
+    window.addEventListener("input", debounce(idJump, 500));
+}
```
FritzOnFire: This is a lot of logic living in global space, either move it to a class, or execute it somewhere else.

index.html
```diff
@@ -1,13 +1,27 @@
...
-
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Onboarding JavaScript Task</title>
+        <script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+        <link rel="stylesheet" href="style.css" />
+        <script src="app.js" defer></script>
```
FritzOnFire: You really should not be using `defer` in the project, but there are no rules against it, so I'm not going to ask you to remove it.

style.css
```diff
@@ -0,0 +1,115 @@
...
+  cursor: pointer;
+  transition: all 0.3s;
+  transition-timing-function: ease-in-out;
+}
+button:hover {
+  letter-spacing: 1px;
+  color: #fff;
+  background-color: #000;
+  border: 1px solid #fff;
+}
```
FritzOnFire: Nice touch :)

FritzOnFire (changes requested):
Please change your spaces to tabs

style.css
```diff
@@ -0,0 +1,115 @@
+* {
+  padding: 0;
```
FritzOnFire: Please convert the spaces to tabs on this file.

app.ts
FritzOnFire: bump

app.ts
```diff
@@ -0,0 +1,292 @@
...
+	const prev = new Prev(); // prev class
+	window.addEventListener("input", debounce(idJump, 500));
+	window.addEventListener("resize", debounce(resizing, 500));
+	getRecordCount()
+		.then(() => {
+			return getHeadings();
+		})
+		.then(() => {
+			return getTable();
+		});
```
FritzOnFire: The classes and functions that you are using here should be defined before you use them.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+// Table Content
+function createTableContent(contentData: any) {
+	const content: HTMLElement | null = document.getElementById("content");
+
+	let table = `<div id="row-${contentData[0]}" class="rows"></div>`;
+	if (content !== null) {
+		content.innerHTML += table;
+	} else {
+		alert("ERROR");
+	}
```
FritzOnFire: You probably also want to return in this `else` as the rest of your code doesn't really work if the `content` does not exist.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+	const heading: HTMLElement | null = document.getElementById("heading");
+
+	let headings = `<div class="headings" id="headings">${headingData}</div>`;
+	if (heading !== null) {
+		heading.innerHTML += headings;
+	}
+}
+
+// Table Content
+function createTableContent(contentData: any) {
```
FritzOnFire: Please specify the type of `contentData`. If you can use it int a `for-loop` then you at the very least know its an array.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+
+// Heading Row (Getting the columns data)
+function getHeadings(): Promise<void> {
+	return fetch("http://localhost:2050/columns", {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(handleErrors)
+		.then((response: Response) => response.json())
+		.then((data: string) => {
```
FritzOnFire: Not sure how this is a `string` when you have a `for-loop` over it in the next few lines...

app.ts
```diff
@@ -0,0 +1,292 @@
...
+			let headingData = data;
```
FritzOnFire: Just rename `data` to `headingData`, no need for this step

app.ts
```diff
@@ -0,0 +1,292 @@
...
+// Gets the number of rows on the screen
+function getNoOfRows(): number {
+	const height = window.innerHeight;
+	let noOfRows = Math.floor(height / 40);
+	return noOfRows;
+}
+
+//// Functions To Create The HTML
+// Heading Row
+function createHeadingRow(headingData: string) {
```
FritzOnFire: Based on how you are using this function in your code, you should probably rename it to `createHeadingColumn`.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+		pageStats.innerHTML = currentStats;
+	}
+
+	return fetch(`http://localhost:2050/records?from=${fromParameter}&to=${toParameter}`, {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(handleErrors)
+		.then((res: Response) => res.json())
+		.then((data: string) => {
```
FritzOnFire: Not sure how this is a `string` when you have a `for-loop` over it in the next few lines...

app.ts
```diff
@@ -0,0 +1,292 @@
...
+			let contentData = data;
+			for (let content of contentData) {
```
FritzOnFire: You all ready have a variable named `content`, you should call this something else, or rename the previous one to something else.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	return fetch(`http://localhost:2050/records?from=${fromParameter}&to=${toParameter}`, {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(handleErrors)
+		.then((res: Response) => res.json())
+		.then((data: string) => {
+			let contentData = data;
```
FritzOnFire: Just rename `data` to `contentData`, no need to have this step.

app.ts
FritzOnFire: Based on how you are using this function, you should probably rename it to `appendTableContent`, or `addRow`, or `appendTableRow`.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+	const content: HTMLElement | null = document.getElementById("content");
+
+	let table = `<div id="row-${contentData[0]}" class="rows"></div>`;
+	if (content !== null) {
+		content.innerHTML += table;
+	} else {
+		alert("ERROR");
+	}
+
+	let rows: HTMLElement | null = document.getElementById("row-" + contentData[0]);
```
FritzOnFire: Should probablly just call this variable `row`, as it is the element for only one `row`.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+				fromParameter = toParameter - getNoOfRows();
+			}
+
+			this.nextCount = 0;
+
+			getTable();
+		};
+
+		if (this.nextButton) {
+			this.nextButton.addEventListener("click", nextDebounce(next, 500));
```
FritzOnFire: If you are dead set on debouncing this button, then you should rather just call the `next` function here, and put your debounce inside the `next` function on the `getTable` function. Then you can get rid of the `nextCount` logic

app.ts
```diff
@@ -0,0 +1,292 @@
...
+
+	constructor() {
+		this.nextCount = 0;
+		this.nextButton = document.getElementById("next");
+
+		const nextDebounce = (fn: any, delay: number) => {
+			return () => {
+				this.nextCount++;
+				clearTimeout(timer);
+				timer = setTimeout(() => {
```
FritzOnFire: You should add this `timer` (with a different name) to the class, then you are less depended on the global variables.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+				}
+
+				this.prevCount = 0;
+
+				getTable();
+			}
+		};
+
+		if (this.prevButton) {
+			this.prevButton.addEventListener("click", prevDebounce(prev, 500));
```
FritzOnFire: Same comment about the debounce that I made on the `Next` class, should help you get rid of the `prevCount`.

app.ts
```diff
@@ -0,0 +1,292 @@
...
+
+	constructor() {
+		this.prevCount = 0;
+		this.prevButton = document.getElementById("prev");
+
+		const prevDebounce = (fn: any, delay: number) => {
+			return () => {
+				this.prevCount++;
+				clearTimeout(timer);
+				timer = setTimeout(() => {
```
FritzOnFire: Same comment about the `timer` I made in the `Next` class

app.ts
```diff
@@ -0,0 +1,292 @@
...
+
+		if (this.prevButton) {
+			this.prevButton.addEventListener("click", prevDebounce(prev, 500));
+		}
+	}
+}
+
+// ID Jump
+function idJump() {
+	const input: any = document.querySelector("input");
```
FritzOnFire: Please don't use `any`.

style.css
```diff
@@ -0,0 +1,115 @@
...
+  font-size: 1em;
+  font-weight: bold;
+  border: 1px solid #000;
+}
+
+/* Table */
+#content {
+  display: flex;
+  flex-direction: column;
+  height: calc(100vh - 15vh);
```
FritzOnFire: `calc` is bad, don't use `calc`

FritzOnFire (changes requested):

app.ts
```diff
@@ -0,0 +1,292 @@
...
+			}
+		})
+		.catch((error: Error) => {
+			console.log(error);
+		});
+}
+
+// Table Content (Getting the table's data)
+function getTable(): Promise<void> {
+	const content: HTMLElement | null = document.getElementById("content");
```
geraldtivatyi, line 80: I think there is a rule about how to use `document.getElementById` in the Javascript Style Guide.

FritzOnFire: Sorry man, I just saw that, I'm removing that section from the style guide :P Had a discussion with cobus a while back, and we preferred the more pure route.

app.ts
```diff
@@ -0,0 +1,286 @@
+let fromParameter = 0;
+let recordCount: number;
+const el = document.getElementById
```
FritzOnFire: Please remove this.

app.ts
```diff
@@ -0,0 +1,286 @@
...
+	const heading: HTMLElement | null = document.getElementById("heading");
+
+	let headings = `<div class="headings" id="headings">${headingData}</div>`;
+	if (heading !== null) {
+		heading.innerHTML += headings;
+	}
+}
+
+// Table Content
+function appendTableContent(contentData: string | string[]) {
```
FritzOnFire, line 32: The way your code is written it should definitely only be `string[]`

app.ts
geraldtivatyi: No problem!!

app.ts
```diff
@@ -0,0 +1,286 @@
...
+
+// Heading Row (Getting the columns data)
+function getHeadings(): Promise<void> {
+	return fetch("http://localhost:2050/columns", {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(handleErrors)
+		.then((response: Response) => response.json())
+		.then((headingData: string | string[]) => {
```
FritzOnFire: The way your code is written it should definitely only be `string[]`

app.ts
```diff
@@ -0,0 +1,286 @@
...
+		pageStats.innerHTML = currentStats;
+	}
+
+	return fetch(`http://localhost:2050/records?from=${fromParameter}&to=${toParameter}`, {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(handleErrors)
+		.then((res: Response) => res.json())
+		.then((contentData: string | string[]) => {
```
FritzOnFire: They way your code is written... neither of these are right. Based off what you get from the back end, should it not be a 2D array?

app.ts
```diff
@@ -0,0 +1,286 @@
...
+	constructor() {
+		this.nextButton = document.getElementById("next");
+
+		const nextDebounce = (fn: () => void, delay: number) => {
+			return () => {
+				clearTimeout(this.nextTimer);
+				this.nextTimer = setTimeout(() => {
+					fn();
+				}, delay);
+			};
```
FritzOnFire: I think you should rather change this to not return a function, that way you don't have to add that weird `()` on line 189.

app.ts
```diff
@@ -0,0 +1,286 @@
...
+		if (this.nextButton) {
+			this.nextButton.addEventListener("click", next);
+		}
+	}
+}
+
+// Previous
+class Prev {
+	prevButton: HTMLElement | null;
+	prevTimer: number = 0
```

app.ts
```diff
@@ -0,0 +1,286 @@
...
+		this.prevButton = document.getElementById("prev");
+
+		const prevDebounce = (fn: () => void, delay: number) => {
+			return () => {
+				clearTimeout(this.prevTimer);
+				this.prevTimer = setTimeout(() => {
+					fn();
+				}, delay);
+			};
+		};
```
FritzOnFire: I think you should rather change this to not return a function, that way you don't have to have that weird `()` on line 231.

app.ts
```diff
@@ -0,0 +1,286 @@
...
+			toParameter = fromParameter + getNoOfRows();
+
+			let end = fromParameter + getNoOfRows();
+
+			if (end > maxRecordsID) {
+				toParameter = maxRecordsID;
+				fromParameter = toParameter - getNoOfRows();
+			}
+
+			nextDebounce(getTable, 500)()
```
FritzOnFire: missing `;`

app.ts
```diff
@@ -0,0 +1,286 @@
...
+		fromParameter = toParameter - getNoOfRows();
+	}
+	getTable();
+}
+
+//// Navigation
+// Next
+class Next {
+	nextButton: HTMLElement | null;
+	nextTimer: number = 0
```
FritzOnFire: missing `;`

app.ts
```diff
@@ -0,0 +1,286 @@
...
+
+				let intOne = fromParameter - prevAmount;
+
+				if (intOne < 0) {
+					fromParameter = 0;
+				} else {
+					fromParameter = intOne;
+				}
+
+				prevDebounce(getTable, 500)()
```

FritzOnFire (changes requested):
Some small changes, but I think you are done.

PR #45, opened 2022-04-12
Ashton Martin

AshtonMar:
Onboarding JS Project

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -3,54 +3,54 @@ let recordNav: any = document.querySelector("#record-navigation-container"); //N
 let headingColumns: any = document.querySelector("#column-headings-container"); //Headings;
 let infoColumns: any = document.querySelector("#info-columns-container"); //Information;
 
+// Creates the navigation.
 function createNavigation() {
   recordNav.innerHTML = `
-  <div class="next-previous-navigation">
   <div class="next-previous-navigation-btns">
     <button class="previous-records-btn">Previous</button>
     <button class="next-records-btn">Next</button>
     <button class="reset-records-btn">First Page</button>
     <button class="last-records-btn">Last Page</button>
   </div>
-</div>
-<div class="current-page-container">
-<p class=current-page></p>
-</div>
-<div class="select-confirm">
+  <div class="select-confirm">
   <select name="record-selection" id="record-selection">
+  <option value="single">Single Records</option>
     <option value="multiple">Multiple Records</option>
-    <option value="single">Single Records</option>
   </select>
   <button onclick="recordSelection()" id="confirmation-btn">
     Confirm Selection
   </button>
-</div>
+  </div>
+  <div class="current-page-container">
+    <p class=current-page></p>
+  </div>
+
   `;
 }
 
+// Heading and info grid being created.
 function createHeadingGrid(headings: any) {
   let headingsData: any = `<h1 class="column-heading">${headings}</h1>`;
   headingColumns.innerHTML += headingsData;
 }
 
-function createGrid(columnData: string) {
-  let infoRowColumns: any = `
-    <div class="info-data-row">
-    <p class="info-column">${columnData[0]}</p>
-    <p class="info-column">${columnData[1]}</p>
-    <p class="info-column">${columnData[2]}</p>
-    <p class="info-column">${columnData[3]}</p>
-    <p class="info-column">${columnData[4]}</p>
-    <p class="info-column">${columnData[5]}</p>
-    <p class="info-column">${columnData[6]}</p>
-    <p class="info-column">${columnData[7]}</p>
-    <p class="info-column">${columnData[8]}</p>
-    <p class="info-column">${columnData[9]}</p>
-    <p class="info-column">${columnData[10]}</p>
-  </div>`;
-  infoColumns.innerHTML += infoRowColumns;
+function dynamicGrid(columnData: any) {
+  // Creates the row that the info will display and adds it to the infoColumnsArea.
+  let infoDataRow = `<div id="info-row-${columnData[0]}" class="info-rows"></div>`;
+  infoColumns.innerHTML += infoDataRow;
+  // Gets the created rows.
+  let finalInfoDataRow: any = document.querySelector(
+    "#info-row-" + columnData[0] + ".info-rows"
+  );
+
+  // Loops through
+  for (let x = 0; x < columnData.length; x++) {
+    let Div = `<p class="info-row-data">${columnData[x]}</p>`;
+    finalInfoDataRow.innerHTML += Div;
+  }
 }
 
+// Adds data in the heading and info grid.
 function fetchData() {
   // Gets headings array
   fetch("http://localhost:2050/columns", {
@@ -84,7 +84,8 @@ function fetchData() {
         .then((data) => {
           let columnDataList = JSON.parse(data);
           for (let i = 0; i < columnDataList.length; i++) {
-            createGrid(columnDataList[i]);
+            // createGrid(columnDataList[i]);
+            dynamicGrid(columnDataList[i]);
           }
         });
     });
@@ -94,8 +95,112 @@ function fetchData() {
 
 fetchData();
 
+// The functionality to move through an amount of records.
+function nextPrevious(firstId: any, finalId: any) {
+  // Btn's used to move from page to page.
+  let nextBtn: any = document.querySelector(".next-records-btn");
+  let previousBtn: any = document.querySelector(".previous-records-btn");
+
+  // Btn's used to move directly to the last and first page.
+  let firstPageBtn: any = document.querySelector(".reset-records-btn");
+  let lastPageBtn: any = document.querySelector(".last-records-btn");
+
+  // The area where the current page is shown.
+  let currentPage: any = document.querySelector(".current-page");
+
+  // The id's of the records that represent the first and last record of the page.
+  let numberOne: number = firstId;
+  let numberTwo: number = finalId;
+  const amountRecords: number = 15;
+
+  // Dsplays the current page.
+  currentPage.innerHTML = numberOne + " / " + numberTwo + " First Page";
+
+  if (numberOne === 0 && numberTwo === 14) {
+    previousBtn.disabled = true;
+  } else {
+    //pass
+  }
+
+  // The button that takes you to the first page.
+  firstPageBtn.addEventListener("click", () => {
+    let nextNumberOne: number = 0;
+    let nextNumberTwo: number = 14;
+    previousBtn.disabled = true;
+    nextBtn.disabled = false;
+
+    currentPage.innerHTML = "";
+    currentPage.innerHTML =
+      nextNumberOne + " / " + nextNumberTwo + " First Page";
+    numberOne = nextNumberOne;
+    numberTwo = nextNumberTwo;
+    nextPageData(numberOne, numberTwo);
+  });
+
+  // The button that takes you to the last page.
+  lastPageBtn.addEventListener("click", () => {
+    let nextNumberOne: number = 999985;
+    let nextNumberTwo: number = 999999;
+    nextBtn.disabled = true;
+    previousBtn.disabled = false;
+
+    currentPage.innerHTML = "";
+    currentPage.innerHTML =
+      nextNumberOne + " / " + nextNumberTwo + " Last Page";
+    numberOne = nextNumberOne;
+    numberTwo = nextNumberTwo;
+    nextPageData(numberOne, numberTwo);
+  });
+
+  // The button that takes you to the next page.
+  nextBtn.addEventListener("click", () => {
+    let nextNumberOne: number = numberOne + amountRecords;
+    let nextNumberTwo: number = numberTwo + amountRecords;
+    previousBtn.disabled = false;
+
+    if (nextNumberOne === 999985 && nextNumberTwo === 999999) {
+      currentPage.innerHTML = "";
+      currentPage.innerHTML =
+        nextNumberOne + " / " + nextNumberTwo + " Final Page";
+      numberOne = nextNumberOne;
+      numberTwo = nextNumberTwo;
+      nextPageData(numberOne, numberTwo);
+      nextBtn.disabled = true;
+    } else {
+      currentPage.innerHTML = "";
+      currentPage.innerHTML = nextNumberOne + " / " + nextNumberTwo;
+      numberOne = nextNumberOne;
+      numberTwo = nextNumberTwo;
+      nextPageData(numberOne, numberTwo);
+    }
+  });
+
+  // The button that takes you to the previous page.
+  previousBtn.addEventListener("click", () => {
+    let nextNumberOne: number = numberOne - amountRecords;
+    let nextNumberTwo: number = numberTwo - amountRecords;
+    nextBtn.disabled = false;
+
+    if (nextNumberOne === 0 && nextNumberTwo === 14) {
+      currentPage.innerHTML = "";
+      currentPage.innerHTML =
+        nextNumberOne + " / " + nextNumberTwo + " First Page";
+      numberOne = nextNumberOne;
+      numberTwo = nextNumberTwo;
+      nextPageData(numberOne, numberTwo);
+      previousBtn.disabled = true;
+    } else {
+      currentPage.innerHTML = "";
+      currentPage.innerHTML = nextNumberOne + " / " + nextNumberTwo;
+      numberOne = nextNumberOne;
+      numberTwo = nextNumberTwo;
+      nextPageData(numberOne, numberTwo);
+    }
+  });
+}
+
 function nextPageData(firstNumber: any, secondNumber: any) {
-  // Fetching information
+  // Fetching information that will display when you go to the next or previous page.
   fetch(
     "http://localhost:2050/records?from=" + firstNumber + "&to=" + secondNumber,
     {
@@ -108,100 +213,12 @@ function nextPageData(firstNumber: any, secondNumber: any) {
       let columnDataList = JSON.parse(data);
       infoColumns.innerHTML = "";
       for (let i = 0; i < columnDataList.length; i++) {
-        createGrid(columnDataList[i]);
+        dynamicGrid(columnDataList[i]);
       }
     });
 }
 
-function nextPrevious(firstId: any, finalId: any) {
-  let nextBtn: any = document.querySelector(".next-records-btn");
-  let previousBtn: any = document.querySelector(".previous-records-btn");
-  let resetBtn: any = document.querySelector(".reset-records-btn");
-  let lastPageBtn: any = document.querySelector(".last-records-btn");
-  let currentPage: any = document.querySelector(".current-page");
-
-  let numberOne: number = firstId;
-  let numberTwo: number = finalId;
-  const amountRecords: number = 15;
-
-  currentPage.innerHTML = numberOne + " -> " + numberTwo + " First Page";
-
-  if (numberOne === 0 && numberTwo === 14) {
-    previousBtn.disabled = true;
-  } else {
-    //pass
-  }
-
-  resetBtn.addEventListener("click", () => {
-    let nextNumberOne: number = 0;
-    let nextNumberTwo: number = 14;
-    previousBtn.disabled = true;
-    nextBtn.disabled = false;
-
-    currentPage.innerHTML = "";
-    currentPage.innerHTML = nextNumberOne + " -> " + nextNumberTwo;
-    numberOne = nextNumberOne;
-    numberTwo = nextNumberTwo;
-    nextPageData(numberOne, numberTwo);
-  });
-
-  lastPageBtn.addEventListener("click", () => {
-    let nextNumberOne: number = 999985;
-    let nextNumberTwo: number = 999999;
-    nextBtn.disabled = true;
-    previousBtn.disabled = false;
-
-    currentPage.innerHTML = "";
-    currentPage.innerHTML = nextNumberOne + " -> " + nextNumberTwo;
-    numberOne = nextNumberOne;
-    numberTwo = nextNumberTwo;
-    nextPageData(numberOne, numberTwo);
-  });
-
-  nextBtn.addEventListener("click", () => {
-    let nextNumberOne: number = numberOne + amountRecords;
-    let nextNumberTwo: number = numberTwo + amountRecords;
-    previousBtn.disabled = false;
-
-    if (nextNumberOne === 999985 && nextNumberTwo === 999999) {
-      currentPage.innerHTML = "";
-      currentPage.innerHTML =
-        nextNumberOne + " -> " + nextNumberTwo + " Final Page";
-      numberOne = nextNumberOne;
-      numberTwo = nextNumberTwo;
-      nextPageData(numberOne, numberTwo);
-      nextBtn.disabled = true;
-    } else {
-      currentPage.innerHTML = "";
-      currentPage.innerHTML = nextNumberOne + " -> " + nextNumberTwo;
-      numberOne = nextNumberOne;
-      numberTwo = nextNumberTwo;
-      nextPageData(numberOne, numberTwo);
-    }
-  });
-
-  previousBtn.addEventListener("click", () => {
-    let nextNumberOne: number = numberOne - amountRecords;
-    let nextNumberTwo: number = numberTwo - amountRecords;
-    nextBtn.disabled = false;
-
-    if (nextNumberOne === 0 && nextNumberTwo === 14) {
-      currentPage.innerHTML = "";
-      currentPage.innerHTML =
-        nextNumberOne + " -> " + nextNumberTwo + " First Page";
-      numberOne = nextNumberOne;
-      numberTwo = nextNumberTwo;
-      nextPageData(numberOne, numberTwo);
-      previousBtn.disabled = true;
-    } else {
-      currentPage.innerHTML = "";
-      currentPage.innerHTML = nextNumberOne + " -> " + nextNumberTwo;
-      numberOne = nextNumberOne;
-      numberTwo = nextNumberTwo;
-      nextPageData(numberOne, numberTwo);
-    }
-  });
-}
+// Checks if you looking for a single or multiple records.
 function recordSelection() {
   let selectionArea: any = recordNav;
   let recordSelector: any = document.querySelector("#record-selection");
@@ -212,11 +229,139 @@ function recordSelection() {
     recordSelectionValue === "multiple" ||
     recordSelectionValue === "single"
   ) {
-    if (recordSelectionValue === "multiple") {
+    if (recordSelectionValue === "single") {
+      let multipleSingleRecordSelection = `
+    <button id="return-btn">Return</button>
+      <div id="user-input-data">
+        <div class="navigation-input-area-id" id="id">
+          <label class="record-labels" for="record-id"
+            >Enter record ID :
+          </label>
+          <input
+            type="number"
+            min="0"
+            name="record-id"
+            id="record-id"
+            class="navigation-input"
+            value="0"
+          />
+        </div>
+      </div>
+    <button id="get-record-btn">See Record</button>
+    `;
+      selectionArea.innerHTML = "";
+      selectionArea.innerHTML = multipleSingleRecordSelection;
+
+      let returnBtn: any = document.querySelector("#return-btn");
+
+      returnBtn.addEventListener("click", () => {
+        let recordsViewed: any;
+        let viewedRecords: any = [];
+        for (let i = 0; i < records.length; i++) {
+          recordsViewed = JSON.parse(records[i])["ID"];
+          viewedRecords.push(recordsViewed);
+        }
+        alert("Records you have viewed " + viewedRecords);
+        headingColumns.innerHTML = "";
+        infoColumns.innerHTML = "";
+        fetchData();
+        createNavigation();
+      });
+
+      let getSingleRecord: any = document.querySelector("#get-record-btn");
+
+      getSingleRecord.addEventListener("click", () => {
+        infoColumns.innerHTML = "";
+        // Id input
+        let IdSelection: any = document.querySelector("#record-id");
+
+        //Values Needed
+        let IdValue = IdSelection.value;
+        let numberCheck: number = 12 - Number(IdValue);
+
+        // Value checks
+        if (IdValue.length === 0) {
+          alert("You have entered an incorrect character.");
+        } else {
+          if (IdValue > 999999) {
+            alert("The record you are looking for does not exist");
+            IdValue = 0;
+          } else {
+            //pass
+          }
+
+          if (typeof numberCheck === "number") {
+            let record = {
+              ID: IdValue,
+            };
+
+            let arrayRecord = JSON.stringify(record);
+            let isInArray = records.includes(arrayRecord);
+            let finalSingleId: number;
+            let properId: any;
+
+            if (IdValue > 999984) {
+              properId = IdValue;
+              finalSingleId = 999999;
+              IdValue = 999984;
+            } else {
+              finalSingleId = Number(IdValue) + 15;
+            }
+
+            if (isInArray === false || isInArray === true) {
+              if (isInArray === true) {
+                let removeIndex = records.indexOf(record["ID"]);
+                records.splice(removeIndex);
+              } else {
+                //pass
+              }
+
+              // Fetching information
+              fetch(
+                "http://localhost:2050/records?from=" +
+                  IdValue +
+                  "&to=" +
+                  finalSingleId,
+                {
+                  method: "GET",
+                  headers: { "Content-Type": "application/json" },
+                }
+              )
+                .then((response) => response.text())
+                .then((data) => {
+                  let columnDataList = JSON.parse(data);
+
+                  infoColumns.innerHTML += "";
+                  for (let i = 0; i < columnDataList.length; i++) {
+                    if (IdValue == 999984) {
+                      IdValue = properId;
+                      console.log(properId);
+                    } else {
+                      //pass
+                    }
+                    createGridSingle(columnDataList[i], IdValue);
+                  }
+                });
+
+              records.push(arrayRecord);
+
+              for (let i = 0; i < records.length; i++) {
+                JSON.parse(records[i]);
+              }
+            } else {
+              alert("There is a problem accessing the record.");
+            }
+          } else {
+            alert("Enter apropriate inputs please.");
+          }
+        }
+      });
+
+      records = records;
+    } else if (recordSelectionValue === "multiple") {
       let multipleRecordSelection = `
-        <button id="multiple-return-btn" class="return-btn">Return</button>
-        <div id="user-input-data">
-        <div class="navigation-input-area">
+      <button id="return-btn">Return</button>
+      <div id="user-input-data">
           <div class="navigation-input-area-id" id="from-id">
             <label class="record-labels" for="record-id"
               >Enter starting record ID :
@@ -229,31 +374,23 @@ function recordSelection() {
               class="navigation-input"
               value="0"
             />
-          </div>
-          <div class="navigation-input-area-id" id="from-id">
-            <label class="record-labels" for="record-id"
-              >Enter stop record ID :
-            </label>
             <input
-              type="number"
-              min="0"
-              name="record-id"
-              id="to-record-id"
-              class="navigation-input"
-              value="15"
-            />
+            type="number"
+            readonly
+            disabled
+            min="0"
+            name="record-id"
+            id="to-record-id"
+            class="navigation-input"
+          />
           </div>
-        </div>
       </div>
-      <button id="multiple-submit-btn" class="submit-btn" >Get Records</button>
+      <button id="submit-btn">Get Records</button>
       `;
       selectionArea.innerHTML = "";
       selectionArea.innerHTML = multipleRecordSelection;
 
-      let returnBtn: any = document.querySelector(
-        "#multiple-return-btn.return-btn"
-      );
-
+      let returnBtn: any = document.querySelector("#return-btn");
       returnBtn.addEventListener("click", () => {
         records = [];
         headingColumns.innerHTML = "";
@@ -262,10 +399,8 @@ function recordSelection() {
         createNavigation();
       });
 
-      let multipleSubmitBtn: any = document.querySelector(
-        "#multiple-submit-btn.submit-btn"
-      );
-      multipleSubmitBtn.addEventListener("click", () => {
+      let SubmitBtn: any = document.querySelector("#submit-btn");
+      SubmitBtn.addEventListener("click", () => {
         let fromIdSelection: any = document.querySelector(
           "#from-record-id.navigation-input"
         );
@@ -274,217 +409,81 @@ function recordSelection() {
         );
 
         let fromIdValue = fromIdSelection.value;
-        let toIdValue = toIdSelection.value;
 
-        if (fromIdValue.length === 0 || toIdValue === 0) {
-          alert(
-            "You left one of the record id inputs empty or you have entered an incorrect character."
-          );
-        } else {
-          if (
-            (fromIdValue !== null && toIdValue !== null) ||
-            (fromIdValue !== "" && toIdValue !== "")
-          ) {
-            const totalRecordsAllowed = 16;
-            let recordCount: number =
-              Number(toIdValue) - Number(fromIdValue) + 1;
-            if (
-              recordCount > totalRecordsAllowed ||
-              typeof recordCount !== "number"
-            ) {
-              let excessRecords = recordCount - totalRecordsAllowed;
-              toIdValue = toIdValue - excessRecords;
-              alert(
-                "The data you entered is incorrect or you trying to access to much records only " +
-                  totalRecordsAllowed +
-                  " records can be accessed at a time. You have " +
-                  excessRecords +
-                  " excess records"
-              );
-              // Fetching information
-              fetch(
-                "http://localhost:2050/records?from=" +
-                  fromIdValue +
-                  "&to=" +
-                  toIdValue,
-                {
-                  method: "GET",
-                  headers: { "Content-Type": "application/json" },
-                }
-              )
-                .then((response) => response.text())
-                .then((data) => {
-                  let columnDataList = JSON.parse(data);
-
-                  infoColumns.innerHTML = "";
-                });
-            } else if (recordCount <= 16 || typeof recordCount === "number") {
-              // Fetching information
-              fetch(
-                "http://localhost:2050/records?from=" +
-                  fromIdValue +
-                  "&to=" +
-                  toIdValue,
-                {
-                  method: "GET",
-                  headers: { "Content-Type": "application/json" },
-                }
-              )
-                .then((response) => response.text())
-                .then((data) => {
-                  let columnDataList = JSON.parse(data);
-
-                  infoColumns.innerHTML = "";
-                  for (let i = 0; i < columnDataList.length; i++) {}
-                });
-            }
-          } else {
-            alert(
-              "Sorry there has been a problem. Please check your inputs make sure you have both filled."
-            );
-          }
-        }
-      });
-    } else if (recordSelectionValue === "single") {
-      let multipleSingleRecordSelection = `
-      <button id="multiple-single-return-btn" class="return-btn">Return</button>
-      <div id="user-input-data">
-        <div class="navigation-input-area">
-          <div class="navigation-input-area-id" id="id">
-            <label class="record-labels" for="record-id"
-              >Enter record ID :
-            </label>
-            <input
-              type="number"
-              min="0"
-              name="record-id"
-              id="record-id"
-              class="navigation-input"
-              value="0"
-            />
-          </div>
-          <div id="multiple-single-selection" class="navigation-input-area-id">
-            <label class="record-labels" for="record-id"
-              >Enter record content letter (Optional) :
-            </label>
-            <select
-              name="record-content"
-              class="navigation-input"
-              id="record-content"
-            >
-              <option value="none">Select Letter</option>
-              <option value="a">A</option>
-              <option value="b">B</option>
-              <option value="c">C</option>
-              <option value="d">D</option>
-              <option value="e">E</option>
-              <option value="f">F</option>
-              <option value="g">G</option>
-              <option value="h">H</option>
-              <option value="i">I</option>
-              <option value="j">J</option>
-            </select>
-          </div>
-        </div>
-      </div>
-      <button id="add-records-btn" class="add-btn">See Records</button>
-      `;
-      selectionArea.innerHTML = "";
-      selectionArea.innerHTML = multipleSingleRecordSelection;
-      let returnBtn: any = document.querySelector(
-        "#multiple-single-return-btn.return-btn"
-      );
-      returnBtn.addEventListener("click", () => {
-        alert("Records you have viewed " + records);
-        headingColumns.innerHTML = "";
-        infoColumns.innerHTML = "";
-        fetchData();
-        createNavigation();
-      });
-      let addMulitpleSingularRecords: any = document.querySelector(
-        "#add-records-btn.add-btn"
-      );
-
-      addMulitpleSingularRecords.addEventListener("click", () => {
-        // Inputs and selects
-        let IdSelection: any = document.querySelector(
-          "#record-id.navigation-input"
-        );
-
-        let letterValueSelection: any =
-          document.querySelector("#record-content");
-
-        //Values Needed
-        let IdValue = IdSelection.value;
-        let letterValue = letterValueSelection.value;
-        let numberCheck: number = 12 - Number(IdValue);
-        const recordAmount: number = 16;
-        let amountOfRecords = records.length;
-
-        if (amountOfRecords > recordAmount) {
-          alert("That is the total amount of records that can be added");
-        } else if (amountOfRecords === 0) {
-          infoColumns.innerHTML = "";
+        if (fromIdValue > 999984 && fromIdValue <= 999999) {
+          fromIdValue = 999984;
+          alert("You have reached the final page");
         } else {
           //pass
         }
 
-        if (IdValue.length === 0 || letterValue === "none") {
-          alert(
-            "You left one of the record id inputs empty or you have entered an incorrect character."
-          );
+        const totalRecordsAllowed = 15;
+
+        let toIdSelectionValue = (toIdSelection.value =
+          Number(fromIdValue) + 15);
+        let toIdValue = toIdSelectionValue;
+
+        let recordCount: number = Number(toIdValue) - Number(fromIdValue);
+
+        if (recordCount !== 15 || fromIdValue > 999999) {
+          fromIdSelection.value = "0";
+          toIdSelection.value = "14";
+          alert("Your entered ID is not correct or does not exist");
         } else {
-          if (
-            typeof numberCheck === "number" &&
-            typeof letterValue === "string"
-          ) {
-            let record = {
-              Id: IdValue,
-              letterId: letterValue,
-            };
-            let arrayRecord = JSON.stringify(record);
-            let isInArray = records.includes(arrayRecord);
-
-            if (isInArray === false) {
-              // Fetching information
-              fetch(
-                "http://localhost:2050/records?from=" +
-                  IdValue +
-                  "&to=" +
-                  IdValue,
-                {
-                  method: "GET",
-                  headers: { "Content-Type": "application/json" },
-                }
-              )
-                .then((response) => response.text())
-                .then((data) => {
-                  let columnDataList = JSON.parse(data);
-
-                  infoColumns.innerHTML += "";
-                  console.log(columnDataList);
-                });
-              records.push(arrayRecord);
-
-              for (let i = 0; i < records.length; i++) {
-                JSON.parse(records[i]);
+          if (totalRecordsAllowed === recordCount && fromIdValue <= 999999) {
+            // Fetching information
+            fetch(
+              "http://localhost:2050/records?from=" +
+                fromIdValue +
+                "&to=" +
+                toIdValue,
+              {
+                method: "GET",
+                headers: { "Content-Type": "application/json" },
               }
-            } else {
-              alert("The record was already added");
-            }
+            )
+              .then((response) => response.text())
+              .then((data) => {
+                let columnDataList = JSON.parse(data);
+
+                infoColumns.innerHTML = "";
+                for (let i = 0; i < columnDataList.length; i++) {
+                  dynamicGrid(columnDataList[i]);
+                }
+              });
           } else {
-            alert("Enter apropriate inputs please");
+            alert("Sorry there has been a problem. Please check your inputs.");
           }
         }
       });
-      records = records;
     } else {
-      alert(
-        "Sorry there has been a problem. The page will reload can you please try again."
-      );
+      alert("Sorry there has been a problem.");
       window.location.reload();
     }
   } else {
-    alert("Error");
+    alert("Something went wrong.");
+  }
+}
+
+// Highlights the record you are searching for.
+function createGridSingle(columnData: string, highlightedRecord: any) {
+  // Creates the row that the info will display and adds it to the infoColumnsArea.
+  let infoDataRow = `<div id="info-row-${columnData[0]}" class="info-rows"></div>`;
+  infoColumns.innerHTML += infoDataRow;
+
+  // Gets the created rows.
+  let finalInfoDataRow: any = document.querySelector(
+    "#info-row-" + columnData[0] + ".info-rows"
+  );
+
+  // Loops through the data.
+  for (let x = 0; x < columnData.length; x++) {
+    let Div = `<p class="info-row-data">${columnData[x]}</p>`;
+    finalInfoDataRow.innerHTML += Div;
+  }
+  if (highlightedRecord == columnData[0]) {
+    finalInfoDataRow.style.color = "red";
+  } else {
+    //pass
   }
 }
--- a/app.ts
+++ b/app.ts
@@ -1,43 +1,44 @@
-let headingColumns: any = document.querySelector("#column-headings-container"); // Headings
-let infoColumns: any = document.querySelector("#info-columns-container"); // Information
 let fromNumber = 0;
 let recordNumberTotal: number;
 let count = 0;
 let timeout = 0;
 
 window.onload = () => {
-  fromNumber = 0;
+    fromNumber = 0;
 };
 
 let debounce = (func: any, delay: number) => {
-  return function () {
-    clearTimeout(timeout);
+    return function () {
+        clearTimeout(timeout);
 
-    timeout = setTimeout(() => {
-      func();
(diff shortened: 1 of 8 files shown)


app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+      if (firstId === 0) {
+        previousBtn.disabled = true;
+        nextBtn.disabled = false;
+      } else if (finalId === 999999) {
+        nextBtn.disabled = true;
+        previousBtn.disabled = false;
+      } else {
+        //pass
+      }
```
anzelIMQS: Remove the else

app.ts
```diff
@@ -0,0 +1,414 @@
...
+          finalId = firstId + amountRecord;
+        } else if (window.innerHeight <= 350 && window.innerHeight > 170) {
+          amountRecord = 2;
+          finalId = firstId + amountRecord;
+        } else if (window.innerHeight <= 170) {
+          amountRecord = 1;
+          finalId = firstId + amountRecord;
+        } else {
+          //pass
+        }
```
anzelIMQS: Remove the else

app.ts
```diff
@@ -0,0 +1,414 @@
...
+        finalId = firstId + 14;
+      } else if (window.innerHeight <= 500 && window.innerHeight > 350) {
+        finalId = firstId + 10;
+      } else if (window.innerHeight <= 350 && window.innerHeight > 170) {
+        finalId = firstId + 2;
+      } else if (window.innerHeight <= 170) {
+        finalId = firstId + 1;
+      } else {
+        //pass
+      }
```
anzelIMQS: Remove the else

app.ts
```diff
@@ -0,0 +1,414 @@
...
+        firstId = finalId - 14;
+      } else if (window.innerHeight <= 500 && window.innerHeight > 350) {
+        firstId = finalId - 10;
+      } else if (window.innerHeight <= 350 && window.innerHeight > 170) {
+        firstId = finalId - 2;
+      } else if (window.innerHeight <= 170) {
+        firstId = finalId - 1;
+      } else {
+        //pass
+      }
```
anzelIMQS: Remove the else.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+      nextBtn.disabled = false;
+      firstId = firstId - (finalId - firstId);
+      finalId = finalId - (finalId - firstId);
+
+      if (firstId < 0) {
+        firstId = 0;
+        previousBtn.disabled = true;
+      } else {
+        //pass
+      }
```
anzelIMQS: Remove the else.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+      if (finalId === 0) {
+        previousBtn.disabled = true;
+      } else {
+        //pass
+      }
```
anzelIMQS: Remove the else.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+  for (let x = 0; x < columnData.length; x++) {
+    let infoData = `<p class="info-row-data">${columnData[x]}</p>`;
+    finalInfoDataRow.innerHTML += infoData;
+  }
+}
+
+function screenSizeCheck() {
+  let fromNumber: any;
+  let toNumber: any;
+  let recordDiff: any;
```
anzelIMQS: Don't use **any** if not necessary. Give them a specific type.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+// Adds data in the heading and info grid.
+function fetchData() {
+  // Gets headings array
+  fetch("http://localhost:2050/columns", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
+    .then((data) => {
```
anzelIMQS: Add a catch to log and handle errors.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+  // Fetching information
+  fetch(
+    "http://localhost:2050/records?from=" + fromNumber + "&to=" + toNumber,
+    {
+      method: "GET",
+      headers: { "Content-Type": "application/json" },
+    }
+  )
+    .then((response) => response.text())
+    .then((data) => {
```
anzelIMQS: Add a catch to log and handle errors.

server/main.go
```diff
@@ -13,11 +13,11 @@ import (
...
 )
 
+var columns = [13]string{"ID", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L"}
 const recordCount = 1000000
-const columnCount = 11
+const columnCount = len(columns); 
 const delayResponse = 500 * time.Millisecond
 
-var columns = [columnCount]string{"ID", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J"}
 var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
anzelIMQS: Don't touch the backend. Frontend only.

anzelIMQS (changes requested):

app.ts
```diff
@@ -0,0 +1,414 @@
+// Get the container from the html where the data will be stored
+let recordNav: any = document.querySelector("#record-navigation-container"); //Navigation area;
```
geraldtivatyi: These do not need to be global variables. Global variables are generally frowned upon. There are a few acceptable global variables but try to remove these from being global.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+  </div>
+  <div class="current-page-container">
+    <p class=current-page></p>
+  </div>
+
+  `;
+}
+
+// The functionality to move through an amount of records.
+function nextPrevious(firstId: any) {
```
geraldtivatyi: Use the `any` type sparingly. Go through the JavaScript Style Guide to see preferences and standards.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+  let currentPage: any = document.querySelector(".current-page");
+  let finalId: any;
+
+  if (recordNav.contains(navigation) === true) {
+    let resizeScreen = () => {
+      let amountRecord: any;
+
+      if (finalId >= 999999) {
+        if (window.innerHeight > 500) {
```
geraldtivatyi: I would avoid doing this check.Try to fit as much readable rows on the screen as possible. Screen sizes will vary so rather try to have a function that calculates and returns the number of rows you can fit on the page and then use that value to set your parameters for fetching the table data.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+          finalId = 999999;
+          firstId = finalId - 14;
+        } else if (window.innerHeight <= 500 && window.innerHeight > 350) {
+          finalId = 999999;
+          firstId = finalId - 10;
+        } else if (window.innerHeight <= 350 && window.innerHeight > 170) {
+          finalId = 999999;
+          firstId = finalId - 2;
+        } else if (window.innerHeight <= 170) {
+          finalId = 999999;
```
geraldtivatyi: `finalId` is being repeated multiple times in the if statements. This is a good sign to either create a method or just simply move that line outside of the if-else statements. Code refactoring basically :P

app.ts
```diff
@@ -0,0 +1,414 @@
...
+          firstId = finalId - 1;
+        } else {
+          //pass
+        }
+      } else {
+        if (window.innerHeight > 500) {
+          amountRecord = 14;
+          finalId = firstId + amountRecord;
```
geraldtivatyi: Some more repetitive code here.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+        currentPage.innerHTML =
+          firstId +
+          " / " +
+          finalId +
+          " " +
+          (finalId - firstId + 1) +
+          " records only.";
+
+        console.log(typeof firstId, typeof finalId);
```
geraldtivatyi: I'm guessing this was for testing your code? You can remove any `console.logs` if they are not necessary.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+      resizeScreen();
+    });
+
+    lastPage.addEventListener("click", () => {
+      previousBtn.disabled = false;
+      nextBtn.disabled = true;
+
+      finalId = 999999;
+
+      if (window.innerHeight > 500) {
```
geraldtivatyi: This checking for the screen size is being done for a second time (similar to earlier in the `nextPrevious()` function), so I think put this logic into its own method which you can call when you need to know whats the number of rows needed for the current screen size.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+  if (typeof fromNumber && typeof toNumber === "undefined") {
+    fromNumber = 0;
+  } else if (typeof fromNumber === "number" && fromNumber !== 0) {
+    console.log(fromNumber);
+    //pass
+  }
+
+  let resizeScreen = () => {
+    if (window.innerHeight > 500) {
```
geraldtivatyi: This window size check is repeated here again.

geraldtivatyi (changes requested):

index.html
```diff
@@ -1,13 +1,18 @@
...
+      charset="utf-8"
+      src="third_party/jquery-2.0.3.min.js"
+    ></script>
+    <link rel="stylesheet" href="./style.css" />
+  </head>
+  <body>
+    <div id="record-navigation-container"></div>
+    <div id="column-headings-container"></div>
+    <div id="info-columns-container"></div>
+    <script src="./app.js"></script>
```
CelesteNaude: Please move your `script` to the `head` tag.

index.html
```diff
@@ -1,13 +1,18 @@
...
-	<p>Hello</p>
-</body>
-
+  <head>
+    <title>JS Onboard Project</title>
+    <script
+      type="text/javascript"
+      charset="utf-8"
+      src="third_party/jquery-2.0.3.min.js"
+    ></script>
```
CelesteNaude: Rather keep this to one line.
```suggestion
<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
```

app.ts
```diff
@@ -0,0 +1,414 @@
+// Get the container from the html where the data will be stored
+let recordNav: any = document.querySelector("#record-navigation-container"); //Navigation area;
+let headingColumns: any = document.querySelector("#column-headings-container"); //Headings;
+let infoColumns: any = document.querySelector("#info-columns-container"); //Information;
```
CelesteNaude: Don't use global variables, rather move them to a class (denounce is fine tho).

app.ts
```diff
@@ -0,0 +1,414 @@
...
+          //pass
+        }
+      }
+
+      if (firstId === 0) {
+        previousBtn.disabled = true;
+        nextBtn.disabled = false;
+      } else if (finalId === 999999) {
+        nextBtn.disabled = true;
+        previousBtn.disabled = false;
```
CelesteNaude: Don't use hard coded values (i.e. 0 and 999999) to do checks on your ids. What would happen if a data set of a different size is given? Try to keep checks on your data as generic as possible.

CelesteNaude (changes requested):

app.ts
```diff
@@ -0,0 +1,414 @@
...
+          " " +
+          (finalId - firstId + 1) +
+          " records only.";
+
+        console.log(typeof firstId, typeof finalId);
+
+        fetchDataScreenSize(firstId, finalId);
+      } else {
+        firstId = firstId;
+        finalId = finalId;
```
anzelIMQS: It surely shouldn't be necessary to set these variables to themselves? Especially since they are global variables in your case.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+
+        console.log(firstId, finalId);
+
+        currentPage.innerHTML =
+          firstId +
+          " / " +
+          finalId +
+          " " +
+          (finalId - firstId + 1) +
+          " records only.";
```
anzelIMQS: Please rather make use of **template literals** here.

It's better suited for string concatenation when you have more than 1 variable in your string.

app.ts
```diff
@@ -0,0 +1,414 @@
...
+      nextBtn.disabled = true;
+
+      finalId = 999999;
+
+      if (window.innerHeight > 500) {
+        firstId = finalId - 14;
+      } else if (window.innerHeight <= 500 && window.innerHeight > 350) {
+        firstId = finalId - 10;
+      } else if (window.innerHeight <= 350 && window.innerHeight > 170) {
+        firstId = finalId - 2;
```
anzelIMQS, line 169: Rather use the subtract and assign operator. e.g. firstId -= 2; 

Please do the same for the others as well.

anzelIMQS (changes requested):
Just a few more little changes.

app.ts
```diff
@@ -0,0 +1,233 @@
...
+
+      nextBtn.disabled = true;
+    }
+
+    fetch("http://localhost:2050/records?from=" + fromNumber + "&to=" + toNumber, {
+      method: "GET",
+      headers: { "Content-Type": "application/json" },
+    })
+      .then((response) => response.text())
+      .then((data) => {
```
anzelIMQS: Implement the catch on the Promise. **.then().catch( () => {});**

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  };
+};
+
+let tryCatch = (func: any) => {
+  try {
+    func;
+  } catch (error) {
+    console.log(error);
+  }
+};
```
anzelIMQS: Let's not do it this way but rather on the Promise.

server/main.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const recordCount = 1000000
-const columnCount = 11
+const columnCount = 11
```
anzelIMQS: Remove the whitespace

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  window.addEventListener("resize", resizeScreen);
+  tryCatch(resizeScreen());
+  tryCatch(nextPrevious(fromNumber));
+}
+
+headingRowCreation();
+
+function recordSelection() {
+  let selectionArea: any = document.querySelector("#record-navigation-container");
+  let SingleRecordSelection = `
```
anzelIMQS: Please make it lowerCamelCase

app.ts
```diff
@@ -0,0 +1,233 @@
...
+        let columnDataList = JSON.parse(data);
+        infoColumns.innerHTML = "";
+        for (let i = 0; i < columnDataList.length; i++) {
+          dynamicGrid(columnDataList[i]);
+        }
+      });
+
+    let currentPage: any = document.querySelector(".current-page");
+    currentPage.innerHTML = "";
+    currentPage.innerHTML = currentPage.innerHTML = fromNumber + " / " + toNumber + " " + (Number(numberOfRows) + 1) + " records only.";
```
anzelIMQS: No need to clear it. If you set it a value, the previous values are overwritten.

app.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+let nextPrevious = (fromNumber: number) => {
+  let numberOfRows = Math.floor(window.innerHeight / 50);
+  let count: number = 0;
+
+  let nextBtn: any = document.querySelector(".next-records-btn");
+  let nextPage = () => {
+    fromNumber = fromNumber + numberOfRows * count;
+    tryCatch(dataRowCreation(fromNumber));
```
anzelIMQS: Remove tryCatch

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  nextPage = debounce(nextPage, 500);
+  nextBtn.addEventListener("click", () => {
+    count++;
+    nextPage();
+  });
+
+  let previousBtn: any = document.querySelector(".previous-records-btn");
+  let previousPage = () => {
+    fromNumber = fromNumber - numberOfRows * count;
+    tryCatch(dataRowCreation(fromNumber));
```
anzelIMQS: Remove tryCatch

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  let firstBtn: any = document.querySelector(".first-page-btn");
+  let firstPage = () => {
+    fromNumber = 0;
+    tryCatch(dataRowCreation(fromNumber));
+  };
+
+  firstPage = debounce(firstPage, 500);
+  firstBtn.addEventListener("click", () => {
+    firstPage();
+  });
```
anzelIMQS: You can pass debounce(firstPage, 500); in directly in this case.

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  let lastPage = () => {
+    let toNumber = 999999;
+    fromNumber = toNumber - numberOfRows;
+    tryCatch(dataRowCreation(fromNumber));
+  };
+
+  lastPage = debounce(lastPage, 500);
+  lastBtn.addEventListener("click", () => {
+    lastPage();
+  });
```
anzelIMQS: You can pass debounce(lastPage, 500); in directly in this case.

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  // Loops through
+  for (let x = 0; x < columnData.length; x++) {
+    let infoData = `<p class="info-row-data">${columnData[x]}</p>`;
+    finalInfoDataRow.innerHTML += infoData;
+  }
+}
+
+let nextPrevious = (fromNumber: number) => {
+  let numberOfRows = Math.floor(window.innerHeight / 50);
+  let count: number = 0;
```
anzelIMQS: Remove count. No need to multiply. The next function is supposed to be executed every time it passes debouncing.

app.ts
```diff
@@ -0,0 +1,233 @@
...
+  let finalInfoDataRow: any = document.querySelector("#info-row-" + columnData[0] + ".info-rows");
+
+  // Loops through
+  for (let x = 0; x < columnData.length; x++) {
+    let infoData = `<p class="info-row-data">${columnData[x]}</p>`;
+    finalInfoDataRow.innerHTML += infoData;
+  }
+}
+
+let nextPrevious = (fromNumber: number) => {
```
anzelIMQS: Probably give this function a more descriptive name

app.ts
```diff
@@ -0,0 +1,233 @@
...
+    fromNumber = fromNumber - numberOfRows * count;
+    tryCatch(dataRowCreation(fromNumber));
+    count = 0;
+  };
+
+  previousPage = debounce(previousPage, 500);
+  previousBtn.addEventListener("click", () => {
+    count++;
+    previousPage();
+  });
```
anzelIMQS: Pass in directly.

app.ts
```diff
@@ -0,0 +1,233 @@
...
+    fromNumber = fromNumber + numberOfRows * count;
+    tryCatch(dataRowCreation(fromNumber));
+    count = 0;
+  };
+
+  nextPage = debounce(nextPage, 500);
+  nextBtn.addEventListener("click", () => {
+    count++;
+    nextPage();
+  });
```
anzelIMQS: Pass in directly

app.ts
```diff
@@ -0,0 +1,233 @@
+let headingColumns: any = document.querySelector("#column-headings-container"); //Headings;
+let infoColumns: any = document.querySelector("#info-columns-container"); //Information;
```
anzelIMQS: Add whitespace between // and comment. Can probably remove the semi-colon as well.

anzelIMQS (changes requested):
Do check some of the comments I made and see if they are applicable somewhere else where I might have missed them.

Please do ask me if something is unclear in what I said.

app.ts
```diff
@@ -0,0 +1,261 @@
...
+  let numberOfRows = Math.floor(window.innerHeight / 50);
+  fromNumber = toNumber - numberOfRows;
+  nextBtn.disabled = true;
+  previousBtn.disabled = false;
+
+  getRecords(fromNumber, toNumber);
+};
+
+lastPageBtn.addEventListener("click", () => {
+  console.log("Hello");
```
anzelIMQS: Remove fancy logging

app.ts
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    toNumber = fromNumber + numberOfRows;
+
+    getRecords(fromNumber, toNumber);
+  }
+};
+
+function getRecords(fromNumber: number, toNumber: number) {
+  fetch("http://localhost:2050/records?from=" + fromNumber + "&to=" + toNumber, {
```
anzelIMQS: Let's make this Template literals.

app.ts
```diff
@@ -0,0 +1,261 @@
...
+    let toNumber = fromNumber + numberOfRows;
+    let check = ["undefined", "string", ""];
+
+    if (check.includes(typeof fromNumber) || fromNumber < 0) {
+      alert("Does not exists");
+      recordIdValue.value = "0";
+    } else if (typeof fromNumber === "number" && fromNumber >= 0) {
+      getRecords(fromNumber, toNumber);
+    } else {
+      alert("Error");
```
anzelIMQS: Tell me what's wrong? Error what?

app.ts
```diff
@@ -0,0 +1,261 @@
...
+            value="0"
+          />
+        </div>
+        <p class="amount-of-records"></p>
+      </div>
+    <button id="get-record-btn">See Record</button>
+    `;
+
+  selectionArea.innerHTML = "";
+  selectionArea.innerHTML = singleRecordSelection;
```
anzelIMQS: Probably not necessary to clear it since you are anyway assigning it a new value that overwrites the previous one?

app.ts
```diff
@@ -0,0 +1,261 @@
...
+  let toNumber: number;
+  let selectionArea: any = document.querySelector("#record-navigation-container");
+
+  if (selectionArea.contains(document.querySelector(".navigation-btns"))) {
+    let numberOfRows = Math.floor(window.innerHeight / 50);
+
+    nextBtn.disabled = false;
+    previousBtn.disabled = false;
+
+    if (fromNumber + numberOfRows >= 999999) {
```
anzelIMQS: I only noticed now you are working on a hardcoded value for the max recordCount (999999). Please ask the API once for that value and store it. You can then use it accordingly instead of the working of a hardcoded values.

server/main.go
```diff
@@ -12,11 +12,9 @@ import (
 	"strconv"
 	"time"
 )
-
 const recordCount = 1000000
 const columnCount = 11
 const delayResponse = 500 * time.Millisecond
-
 var columns = [columnCount]string{"ID", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J"}
```
anzelIMQS: There should be any changes to this file. Looks like you accidentally removed the new lines. Make sure this file doesn't even popup on Files changed on GitHub.

anzelIMQS (changes requested):
Make sure to make use of the API recordCount. Also, remove some redundancy.

app.ts
```diff
@@ -0,0 +1,287 @@
+let headingColumns: any = document.querySelector("#column-headings-container"); // Headings
+let infoColumns: any = document.querySelector("#info-columns-container"); // Information
```
FritzOnFire: The way your code is written these lines will execute before those elements actually exist on the webpage. So they can definitely not be here. You should only try and get things from the webpage after it has been rendered.

FritzOnFire: Also change the `any` to the actual type.

server/main.go
```diff
@@ -12,7 +12,6 @@ import (
 	"strconv"
 	"time"
 )
-
```
FritzOnFire: This file should not show up in this PR.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+
+let createNavigation = () => {
+  let recordNav: any = document.querySelector("#record-navigation-container"); // Navigation area
+  recordNav.innerHTML = `
+  <div class="navigation-btns">
+    <button value="first" class="first-page-btn">First Page</button>  
+    <button value="previous" class="previous-records-btn">Previous</button>
+    <button value="next" class="next-records-btn">Next</button>
+    <button value="last" class="last-page-btn">Last Page</button>  
+    <button onclick="recordSelection()" id="confirmation-btn">Get Record</button>
```
FritzOnFire: You should not specify your `onclick` in the HTML. You should do it in the typescript code.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+  return function () {
+    clearTimeout(timeout);
+
+    timeout = setTimeout(() => {
+      func();
+    }, delay);
+  };
+};
+
+let createNavigation = () => {
```
FritzOnFire: Please change this variable to a normal function, seeing as that you never override the value.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+      <div id="user-input-data">
+        <div class="navigation-input-area-id" id="id">
+          <label class="record-labels" for="record-id"
+            >Enter record ID :
+          </label>
+          <input
+            type="text"
+            min="0"
+            minlength="1"
+            maxlength="6"
```
FritzOnFire: Hardcoding the max length here is a bad idea, as the maximum amount of rows that you could get is unknown.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    `;
+
+  selectionArea.innerHTML = singleRecordSelection;
+
+  let returnBtn: any = document.querySelector("#return-btn");
+  let recordIdValue: any = document.querySelector("#record-id.navigation-input");
+  let numberOfRows = Math.floor(window.innerHeight / 50);
+
+  returnBtn.addEventListener("click", () => {
+    window.location.reload();
```
FritzOnFire: you definitely do not want this line anywhere in your project... This is a single page app, reloading is not an option.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    let finalRecord = recordNumberTotal - 1;
+
+    if (toNumber > finalRecord) {
+      toNumber = finalRecord;
+      fromNumber = toNumber - numberOfRows;
+    }
+
+    let check = ["undefined", "string", ""];
+
+    try {
```
FritzOnFire: You should be able to complete this project with out using any `try-catches` (aside from the catch that comes with a promise of course).

app.ts
```diff
@@ -0,0 +1,287 @@
...
+  });
+}
+
+function createHeadingGrid(headings: any) {
+  let headingsData: any = `<h1 class="column-heading">${headings}</h1>`;
+  headingColumns.innerHTML += headingsData;
+}
+
+let recordCount = () => {
+  fetch("http://localhost:2050/recordCount", {
```
FritzOnFire: You should add a `return` in front of the `fetch`. Then another function should be able to wait on this function.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+  });
+}
+
+function createHeadingGrid(headings: any) {
+  let headingsData: any = `<h1 class="column-heading">${headings}</h1>`;
+  headingColumns.innerHTML += headingsData;
+}
+
+let recordCount = () => {
```
FritzOnFire: Please change this to a normal function, seeing as that you are not overriding the variable anywhere.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
+    .then((data) => {
+      recordNumberTotal = JSON.parse(data);
+    })
+    .catch((error) => {
+      console.log(error);
+    });
```
FritzOnFire: Please remove the `()` around the parameters or specify the types, as the compiler currently thinks they are of type `any`.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+  let headingsData: any = `<h1 class="column-heading">${headings}</h1>`;
+  headingColumns.innerHTML += headingsData;
+}
+
+let recordCount = () => {
+  fetch("http://localhost:2050/recordCount", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
```
FritzOnFire: You should check if `response` has an error here.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    .then((data) => {
+      recordNumberTotal = JSON.parse(data);
+    })
+    .catch((error) => {
+      console.log(error);
+    });
+};
+
+let headingRowCreation = () => {
```
FritzOnFire: Please change this variable into a normal function, as you don't seem to be overriding the variable anywhere.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+  fetch("http://localhost:2050/columns", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
+    .then((data) => {
```
FritzOnFire: Please remove the `()` around the parameters or specify the types, as the compiler currently thinks they are of type `any`.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+      console.log(error);
+    });
+};
+
+let headingRowCreation = () => {
+  fetch("http://localhost:2050/columns", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
```
FritzOnFire: You should check if `response` has an error here.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+      let headingDataList = JSON.parse(data);
+      let headings: string;
+
+      for (let i = 0; i < headingDataList.length; i++) {
+        headings = headingDataList[i];
+        createHeadingGrid(headings);
+      }
+      resizeScreenData();
+    })
+    .catch((error) => {
```
FritzOnFire: Please remove the `()` around the `error` parameter or specify the type, as the compiler currently thinks it is of type `any`.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    toNumber = fromNumber + numberOfRows;
+
+    getRecords(fromNumber, toNumber);
+  }
+};
+
+function getRecords(fromNumber: number, toNumber: number) {
+  fetch(`http://localhost:2050/records?from=${fromNumber}&to=${toNumber}`, {
```
FritzOnFire: You should add a `return` in front of your `fetch` to allow other functions to wait on this function.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
+    .then((data) => {
```
FritzOnFire: Please remove the `()` around the parameters or specify the types, as the compiler currently thinks they are of type `any`.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+    getRecords(fromNumber, toNumber);
+  }
+};
+
+function getRecords(fromNumber: number, toNumber: number) {
+  fetch(`http://localhost:2050/records?from=${fromNumber}&to=${toNumber}`, {
+    method: "GET",
+    headers: { "Content-Type": "application/json" },
+  })
+    .then((response) => response.text())
```
FritzOnFire: You should check if `response` has an error here.

app.ts
```diff
@@ -0,0 +1,287 @@
...
+      let columnDataList = JSON.parse(data);
+      infoColumns.innerHTML = "";
+      for (let i = 0; i < columnDataList.length; i++) {
+        dynamicGrid(columnDataList[i]);
+      }
+
+      let currentPage: any = document.querySelector(".current-page");
+      currentPage.innerHTML = `${fromNumber}  / ${toNumber}.`;
+    })
+    .catch((error) => {
```
FritzOnFire: Please remove the `()` around the `error` parameter or specify the type, as the compiler currently thinks it is of type `any`.

FritzOnFire (changes requested):
Please convert your spaces into tabs.

Please remove the `text` file from the PR.

Please remove all the `any` types. The only place I know of that you would need an any is when getting data from the back-end, but most people just specify the actual type there anyway.

You have a lot of global variables and logic being executed in global scope. Please move them into a class, or somewhere where they are no longer considered to be `global`. (Hint, the class option will ensure that you get way less future review changes)

app.ts
```diff
@@ -0,0 +1,312 @@
+let fromNumber = 0;
+let recordNumberTotal: number;
+let count = 0;
+let timeout = 0;
+
+window.onload = () => {
+    fromNumber = 0;
+};
+
+function checkResponseError(response: Response) {
```
FritzOnFire, line 6: Please specify a return type.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    <button id="confirmation-btn">Get Record</button>
+  </div>
+  <div class="current-page-container">
+    <p id=current-page></p>
+  </div>
+  `;
+    }
+}
+
+function getRecords(fromNumber: number, toNumber: number) {
```
FritzOnFire: Please specify a return type

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    return fetch(`http://localhost:2050/records?from=${fromNumber}&to=${toNumber}`, {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(checkResponseError)
+        .then((response: Response) => response.text())
```
FritzOnFire: You should rather use `.json()`. Then you don't have to parse the result and check if it successfully parsed in the next section.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        .then((data) => {
```
FritzOnFire: Either remove the `()` or specify the type, as the compiler currently thinks the type is `any`.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            let columnDataList = JSON.parse(data);
+            let infoColumns: HTMLElement | null = document.getElementById("info-columns-container"); // Information
+
+            console.log("Hello");
+            (infoColumns as HTMLDivElement).innerHTML = "";
```
FritzOnFire: You need to check if `infoColumns` is `null`, and handle it appropriately. Instead of hard casting the value.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(checkResponseError)
+        .then((response: Response) => response.text())
+        .then((data) => {
+            let columnDataList = JSON.parse(data);
+            let infoColumns: HTMLElement | null = document.getElementById("info-columns-container"); // Information
+
+            console.log("Hello");
```
FritzOnFire: Stray console log.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            (infoColumns as HTMLDivElement).innerHTML = "";
+            for (let i = 0; i < columnDataList.length; i++) {
```
FritzOnFire: You should change this to a `for-of` loop seeing as that you are not using the counter.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+                dynamicGrid(columnDataList[i]);
+            }
+
+            let currentPage: HTMLElement | null = document.getElementById("current-page");
+            (currentPage as HTMLParagraphElement).innerHTML = `${fromNumber}  / ${toNumber}.`;
```
FritzOnFire: Same comment as before, you need to check if the value is `null`

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        })
+        .catch((error: Error) => {
+            console.log(error);
+        });
+}
+
+function recordSelection() {
+    let recordNav: HTMLElement | null = document.getElementById("record-navigation-container"); // Navigation area
+    if (recordNav !== null) {
```
FritzOnFire: Rather do a
```ts
if (!recordNav) {
    // Error/Return/Alert what happened.
}
```

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        let singleRecordSelection = `
+  <button id="return-btn">Return</button>
+    <div id="user-input-data">
```
FritzOnFire: You should put the `button` and the `div` on the same indentation, otherwise it looks like the `div` is inside the `button`.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+          type="text"
+          name="record-id"
+          id="record-id"
+          class="navigation-input"
+          value="0"
+        />
+      </div>
+      <p class="amount-of-records"></p>
+    </div>
+  <button id="get-record-btn">See Record</button>
```
FritzOnFire: Same comment as before, but here is looks like your closing a `button`.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+  `;
+
+        recordNav.innerHTML = singleRecordSelection;
+
+        let returnBtn: HTMLElement | null = document.getElementById("return-btn");
+        let recordIdInput: HTMLElement | null = document.getElementById("record-id");
+        let numberOfRows = Math.floor(window.innerHeight / 50);
+        let getSingleRecord: HTMLElement | null = document.getElementById("get-record-btn");
+
+        returnBtn?.addEventListener("click", () => {
```
FritzOnFire: The `?` is nice for when you want to do optional things that don't need to be there. But seeing as that you just made the dom elements, you definitly expect them to be there, so you should check if any of them are `null` and handle the errors accordingly.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            console.log("hello");
```
FritzOnFire: Stray console log.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            createNavigation();
+            fromNumber = 0;
+            let toNumber = fromNumber + numberOfRows;
+            getRecords(fromNumber, toNumber);
```
FritzOnFire: Add a comment saying that you are resetting the grid back to the starting page and remove the `let toNumber` line and just pass the `numberOfRows` as an argument to the `getRecords` function.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        });
+
+        getSingleRecord?.addEventListener("click", () => {
```
FritzOnFire: Same comment about the `?`

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            let recordIdValue = (recordIdInput as HTMLInputElement).value;
```
FritzOnFire: Check for `null`

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            fromNumber = Number(recordIdValue);
+
+            console.log(fromNumber);
```
FritzOnFire: Stray console log

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            }
+        });
+    }
+}
+
+function createHeadingGrid(headings: any) {
+    let headingColumns: HTMLElement | null = document.getElementById("column-headings-container"); // Headings
+    let headingsData = `<h1 class="column-heading">${headings}</h1>`;
+
+    (headingColumns as HTMLDivElement).innerHTML += headingsData;
```
FritzOnFire: Check for `null`

app.ts
```diff
@@ -0,0 +1,312 @@
...
+                alert("Does not exists");
+                recordIdValue = "0";
+            } else if (typeof fromNumber === "number" && fromNumber >= 0) {
+                getRecords(fromNumber, toNumber);
+            }
+        });
+    }
+}
+
+function createHeadingGrid(headings: any) {
```
FritzOnFire: Please specify the real type here.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    let headingColumns: HTMLElement | null = document.getElementById("column-headings-container"); // Headings
+    let headingsData = `<h1 class="column-heading">${headings}</h1>`;
+
+    (headingColumns as HTMLDivElement).innerHTML += headingsData;
+}
+
+function recordCount() {
```
FritzOnFire: Please add a return type

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    return fetch("http://localhost:2050/recordCount", {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(checkResponseError)
+        .then((response: Response) => response.text())
```
FritzOnFire: You should rather use `.json()` to avoid parsing the json in the next section.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        .then((data) => {
```
FritzOnFire: Please remove the `()` or specify the type, as the compiler currently thinks that the type is `any`

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            recordNumberTotal = JSON.parse(data);
+        })
+        .catch((error: Error) => {
+            console.log(error);
+        });
+}
+
+function headingRowCreation() {
```
FritzOnFire: Please specify the return type.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    return fetch("http://localhost:2050/columns", {
+        method: "GET",
+        headers: { "Content-Type": "application/json" },
+    })
+        .then(checkResponseError)
+        .then((response: Response) => response.text())
```
FritzOnFire: You should rather use `.json()` to avoid parsing the json in the next section

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        .then((data) => {
```
FritzOnFire: Please remove the `()` or specify the type, as the compiler currently thinks that the type is `any`

app.ts
```diff
@@ -0,0 +1,312 @@
...
+            let headingDataList = JSON.parse(data);
+            let headings: string;
+
+            for (let i = 0; i < headingDataList.length; i++) {
```
FritzOnFire: Change this to a `for-of` loop.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+                createHeadingGrid(headings);
+            }
+            resizeScreenData();
+        })
+        .catch((error: Error) => {
+            console.log(error);
+        });
+}
+
+function dynamicGrid(columnData: any) {
```
FritzOnFire: Change the type for `columnData` to its actual type.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    let infoColumns: HTMLElement | null = document.getElementById("info-columns-container"); // Information
+    // Creates the row that the info will display and adds it to the infoColumnsArea.
+    let infoDataRow = `<div id="info-row-${columnData[0]}" class="info-rows"></div>`;
+
+    if (infoColumns !== null) {
```
FritzOnFire: You should probably error or something if this happens, otherwise the user will not know whats wrong.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
+
+function resizeScreenData() {
+    let toNumber: number;
+    let recordNav: HTMLElement | null = document.getElementById("record-navigation-container");
+    let nextBtn: HTMLElement | null = document.getElementById("next-records-btn");
+    let previousBtn: HTMLElement | null = document.getElementById("previous-records-btn");
+    let navBtns = document.getElementById("navigation-btns");
+    if (recordNav !== null) {
```
FritzOnFire: You should probably error or something if this happens, otherwise the user will not know whats wrong.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        if (recordNav.contains(navBtns as HTMLDivElement)) {
+            let numberOfRows = Math.floor(window.innerHeight / 50);
+
+            (nextBtn as HTMLButtonElement).disabled = false;
+            (previousBtn as HTMLButtonElement).disabled = false;
```
FritzOnFire: You should check for nulls.

app.ts
```diff
@@ -0,0 +1,312 @@
...
+        fromNumber = finalRecord - numberOfRows;
+        (nextBtn as HTMLButtonElement).disabled = true;
+        (previousBtn as HTMLButtonElement).disabled = false;
+
+        getRecords(fromNumber, finalRecord);
+    };
+    lastPageBtn?.addEventListener("click", () => {
+        lastPage();
+    });
+}
```
FritzOnFire: This is to much logic happening in global space, please either move it into a class, or have it executed somewhere else.

index.html
```diff
@@ -1,13 +1,14 @@
...
-
-<body>
-	<p>Hello</p>
-</body>
-
+  <head>
+    <title>JS Onboard Project</title>
+    <script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+    <link rel="stylesheet" href="./style.css" />
+    <script src="./app.js" defer></script>
```
FritzOnFire: You should not really be using `defer` for this project... but there aren't any rules against it, so I can't ask you to remove it.

server/main.go
FritzOnFire: bump

FritzOnFire (changes requested):
Please change the spaces to tabs.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+};
+
+function checkResponseError(response: Response) {
+	if (!response.ok) {
+		throw Error(response.statusText);
+	}
+	return response;
+}
+
+function debounce(func: any, delay: number) {
```
FritzOnFire: Please specify the type of `func`. Im pretty sure it should be `() => void`.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+}
+
+function getRecords(fromNumber: number, toNumber: number): Promise<void> {
+	return fetch(`http://localhost:2050/records?from=${fromNumber}&to=${toNumber}`, {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(checkResponseError)
+		.then((response: Response) => response.json())
+		.then((data: string) => {
```
FritzOnFire: Not sure how this is a `string`, and then a few lines down you have a `for-loop` over it?

app.ts
```diff
@@ -0,0 +1,319 @@
...
+let recordNumberTotal: number;
+let count = 0;
+let timeout = 0;
+
+window.onload = () => {
+	createNavigation();
+	recordCount();
+	headingRowCreation();
+	fromNumber = 0;
+	new pageNavigation();
```
FritzOnFire: You should create the functions and classes before you use them (They should appear before the `window.onload` line)

app.ts
```diff
@@ -0,0 +1,319 @@
...
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(checkResponseError)
+		.then((response: Response) => response.json())
+		.then((data: string) => {
+			let infoColumns: HTMLElement | null = document.getElementById("info-columns-container"); // Information
+			let currentPage: HTMLElement | null = document.getElementById("current-page");
+
+			if (infoColumns !== null) {
+				(infoColumns as HTMLDivElement).innerHTML = "";
```
FritzOnFire: You don't have to hard cast to `HTMLDivElement` if you have checked for the `null`. Please remove it.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+
+				for (let i of data) {
+					dynamicGrid(i);
+				}
+			}
+
+			if (currentPage !== null) {
+				(currentPage as HTMLParagraphElement).innerHTML = `${fromNumber}  / ${toNumber}.`;
```
FritzOnFire: Same as the comment before, please remove the hard cast.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+		.catch((error: Error) => {
+			console.log(error);
+		});
+}
+
+function recordSelection() {
+	let recordNav: HTMLElement | null = document.getElementById("record-navigation-container"); // Navigation area
+	if (!recordNav) {
+		alert("The navigation is not working correctly refresh the page");
+	} else {
```
FritzOnFire: Please add a `return` after the alert and remove the else, no need to have your whole function inside the else of an if statement.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+		}
+	}
+}
+
+function createHeadingGrid(headings: string) {
+	let headingColumns: HTMLElement | null = document.getElementById("column-headings-container"); // Headings
+	let headingsData = `<h1 class="column-heading">${headings}</h1>`;
+
+	if (headingColumns !== null) {
+		(headingColumns as HTMLDivElement).innerHTML += headingsData;
```
FritzOnFire: No need to hard cast after you have checked the `null`. Please remove it.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+}
+
+function headingRowCreation(): Promise<void> {
+	return fetch("http://localhost:2050/columns", {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(checkResponseError)
+		.then((response: Response) => response.json())
+		.then((data: string) => {
```
FritzOnFire: Not sure how `data` is a `string` when you have a `for-loop` over it on the next line...

app.ts
```diff
@@ -0,0 +1,319 @@
...
+				createHeadingGrid(i);
+			}
+			resizeScreenData();
+		})
+		.catch((error: Error) => {
+			console.log(error);
+		});
+}
+
+function dynamicGrid(columnData: string) {
```
FritzOnFire, line 177: Not sure how `columnData` is a `string` when you have a `for-loop` over it a few lines down...

app.ts
```diff
@@ -0,0 +1,319 @@
...
+	let toNumber: number;
+	let recordNav: HTMLElement | null = document.getElementById("record-navigation-container");
+	let nextBtn: HTMLElement | null = document.getElementById("next-records-btn");
+	let previousBtn: HTMLElement | null = document.getElementById("previous-records-btn");
+	let navBtns = document.getElementById("navigation-btns");
+	if (recordNav !== null) {
+		if (recordNav.contains(navBtns as HTMLDivElement)) {
+			let numberOfRows = Math.floor(window.innerHeight / 50);
+
+			(nextBtn as HTMLButtonElement).disabled = false;
```
FritzOnFire: You should check if `nextBtn` is `null`, instead of hard casting.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+			(previousBtn as HTMLButtonElement).disabled = false;
```
FritzOnFire: You should check if `previousBtn` is `null`, instead of hard casting.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+			}
+
+			toNumber = fromNumber + numberOfRows;
+
+			getRecords(fromNumber, toNumber);
+		}
+	}
+}
+
+window.addEventListener("resize", debounce(resizeScreenData, 500));
```
FritzOnFire: You should not have this is global scope.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+
+		if (this.nextBtn !== null && this.previousBtn !== null && this.firstPageBtn !== null && this.lastPageBtn !== null) {
+			let nextPage = () => {
+				let numberOfRows = Math.floor(window.innerHeight / 50);
+				fromNumber = fromNumber + numberOfRows * count;
+				let toNumber = fromNumber + numberOfRows;
+
+				let finalRecord = recordNumberTotal - 1;
+
+				(this.previousBtn as HTMLButtonElement).disabled = false;
```
FritzOnFire: You have already checked for the `null`, so no need to hard cast. You probably want to change the type of the variable to `HTMLButtonElement` where you defined it to allow this code to keep compiling.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+
+				if (toNumber >= finalRecord) {
+					(this.nextBtn as HTMLButtonElement).disabled = true;
```
FritzOnFire: Same comment as before about not hard casting.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+					fromNumber = finalRecord - numberOfRows;
+				}
+
+				getRecords(fromNumber, toNumber);
+				count = 0;
+			};
+			this.nextBtn.addEventListener("click", () => {
+				count++;
+				nextPage = debounce(nextPage, 500);
+				nextPage();
```
FritzOnFire: You definitely want to remove this line.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+					(this.nextBtn as HTMLButtonElement).disabled = true;
+					fromNumber = finalRecord - numberOfRows;
+				}
+
+				getRecords(fromNumber, toNumber);
+				count = 0;
+			};
+			this.nextBtn.addEventListener("click", () => {
+				count++;
+				nextPage = debounce(nextPage, 500);
```
FritzOnFire: If your deadset on debouncing this functionality, then you should debounce the `getRecords` function call inside `nextPage` instead, that way you can get rid of the `count` variable.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+				}
+
+				getRecords(fromNumber, toNumber);
+				count = 0;
+			};
+			this.previousBtn.addEventListener("click", () => {
+				count++;
+				previousPage = debounce(previousPage, 500);
+				previousPage();
+			});
```
FritzOnFire: Same comments as I made on `nextPage`.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+			toNumber = fromNumber + numberOfRows;
+
+			getRecords(fromNumber, toNumber);
+		}
+	}
+}
+
+window.addEventListener("resize", debounce(resizeScreenData, 500));
+
+class pageNavigation {
```
FritzOnFire: Classes need to be PascalCase

app.ts
```diff
@@ -0,0 +1,319 @@
...
+				previousPage = debounce(previousPage, 500);
+				previousPage();
+			});
+
+			let firstPage = () => {
+				fromNumber = 0;
+				let numberOfRows = Math.floor(window.innerHeight / 50);
+				let toNumber = fromNumber + numberOfRows;
+				(this.nextBtn as HTMLButtonElement).disabled = false;
+				(this.previousBtn as HTMLButtonElement).disabled = true;
```
FritzOnFire: Same comment about the other buttons, you have already checked the `null`, no need to hard cast, just change the type where you defined these buttons.

app.ts
```diff
@@ -0,0 +1,319 @@
...
+			this.firstPageBtn.addEventListener("click", () => {
+				firstPage();
+			});
+
+			let lastPage = () => {
+				let finalRecord = recordNumberTotal - 1;
+				let numberOfRows = Math.floor(window.innerHeight / 50);
+				fromNumber = finalRecord - numberOfRows;
+				(this.nextBtn as HTMLButtonElement).disabled = true;
+				(this.previousBtn as HTMLButtonElement).disabled = false;
```
FritzOnFire: Same as previous comment about these two buttons.

index.html
FritzOnFire: Bump

index.html
```diff
@@ -1,13 +1,18 @@
...
+      src="third_party/jquery-2.0.3.min.js"
+    ></script>
+    <link rel="stylesheet" href="./style.css" />
+  </head>
+  <body>
+    <div id="record-navigation-container"></div>
+    <div id="column-headings-container"></div>
+    <div id="info-columns-container"></div>
+  </body>
+  <script src="./app.js"></script>
```
FritzOnFire: This line needs to remain in your `head` tag.

index.html
```diff
@@ -1,13 +1,18 @@
...
-<head>
-	<title>JS Onboard Project</title>
-	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
-</head>
-
-<body>
-	<p>Hello</p>
-</body>
-
+  <head>
```
FritzOnFire: Please convert this file back to using tabs.

style.css
```diff
@@ -0,0 +1,116 @@
+* {
+    margin: 0;
```
FritzOnFire: Please convert this file to use tabs

FritzOnFire (changes requested):
`debounces` or `throttles` may not exceed 250ms, ideally the closer to 50ms the better. Please change yours, 500ms is way to much.

app.ts
FritzOnFire: Bump

app.ts
```diff
@@ -0,0 +1,336 @@
...
+}
+
+function getRecords(fromNumber: number, toNumber: number): Promise<void> {
+	return fetch(`http://localhost:2050/records?from=${fromNumber}&to=${toNumber}`, {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(checkResponseError)
+		.then((response: Response) => response.json())
+		.then((data: TemplateStringsArray) => {
```
```suggestion
		.then((data: string[][]) => {
```

app.ts
```diff
@@ -0,0 +1,336 @@
...
+	let recordIdInput = <HTMLInputElement | null>document.getElementById("record-id");
+	let numberOfRows = Math.floor(window.innerHeight / 50);
+	let getSingleRecord: HTMLElement | null = document.getElementById("get-record-btn");
+
+	if (returnBtn !== null) {
+		// Resets to the first page
+		returnBtn.addEventListener("click", () => {
+			createNavigation();
+			recordCount();
+			resizeScreenData()
```

app.ts
```diff
@@ -0,0 +1,336 @@
...
+			fromNumber = 0;
+			getRecords(fromNumber, fromNumber + numberOfRows)
```

FritzOnFire: You have no guarentee that `recordCount` will finish before `getRecords` is called. You should change this to only call `getRecords` if `recordCount` has finished.

app.ts
```diff
@@ -0,0 +1,336 @@
...
+	let recordNav: HTMLElement | null = document.getElementById("record-navigation-container");
+	let nextBtn = <HTMLButtonElement | null>document.getElementById("next-records-btn");
+	let previousBtn = <HTMLButtonElement | null>document.getElementById("previous-records-btn");
+	let navBtns = document.getElementById("navigation-btns");
+	if (recordNav !== null) {
+		if (recordNav.contains(navBtns as HTMLDivElement)) {
+			let numberOfRows = Math.floor(window.innerHeight / 50);
+
+			nextBtn!.disabled = false;
+			previousBtn!.disabled = false;
```
FritzOnFire: How do you know if these two buttons are not `null`?

app.ts
```diff
@@ -0,0 +1,336 @@
...
+
+			let finalRecord = recordNumberTotal - 1;
+
+			if (fromNumber + numberOfRows >= finalRecord) {
+				fromNumber = finalRecord - numberOfRows;
+
+				nextBtn!.disabled = true;
+				previousBtn!.disabled = false;
```
FritzOnFire: How do you know if these two buttons are not `null`?

app.ts
```diff
@@ -0,0 +1,336 @@
...
+			} else if (fromNumber <= 0) {
+				fromNumber = 0;
+
+				nextBtn!.disabled = false;
+				previousBtn!.disabled = true;
```
FritzOnFire: How do you know if these two buttons are not `null`?

app.ts
```diff
@@ -0,0 +1,336 @@
...
+}
+
+function headingRowCreation(): Promise<void> {
+	return fetch("http://localhost:2050/columns", {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	})
+		.then(checkResponseError)
+		.then((response: Response) => response.json())
+		.then((data: TemplateStringsArray) => {
```
```suggestion
		.then((data: string[]) => {
```

app.ts
```diff
@@ -0,0 +1,336 @@
...
+
+function navigationDebounce(func: (argOne: number, argTwo: number) => void, delay: number) {
+	return function (argOne: number, argTwo: number) {
+		clearTimeout(timeout);
+
+		timeout = setTimeout(() => {
+			func(argOne, argTwo);
+		}, delay);
+	};
+}
```
FritzOnFire: Please move this into the class. The more things you put in a class the better.

FritzOnFire (changes requested):
Rather never use the `!` to indicate that something is not `null`. It will really bite you in the future and cause a lot of extra work.

PR #46, opened 2023-08-11
Onboarding Project

Ntobe99:
Please Review


index.html
```diff
@@ -3,10 +3,32 @@
...
+			<button id="prevBtn" class="btn">Prev</button>
+			<span id="pageInfo" ></span>
+			<button id="nextBtn" class="btn">Next</button>
+			<div class="search-container">
+				<input type="text" id="fromInput" placeholder="Search ID...">
+				<button id="searchBtn"  class="btn">Search</button>
+			  </div>
+			  
+			</div>	
+	<script src="/app.js"></script>
```
DevinFledermaus: The script tag can be moved into head tag

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		}
+	}
+	// Method to fetch total record count from the server
+	async recordCount() {
+		try {
+			const response = await this.fetchData('http://localhost:2050/recordCount');
+			this.totalItems = response;
+		} catch (error) {
+			console.error('Failed to fetch record count', error);
+			throw error;
```
```suggestion
			throw new Error('Failed to fetch record count', error)
```

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	//fectch column names
+	async fetchColumns() {
+		try {
+			const response = await this.fetchData('http://localhost:2050/columns');
+			const res = JSON.parse(response);
+			this.columnNames = res.map((columnName: any) => ({ name: columnName }));
+			this.data = new Array<GridData>(this.columnNames.length);
+		} catch (error) {
+			console.error('Failed to fetch columns', error);
+			throw error;
```
DevinFledermaus: same as above

app.ts
```diff
@@ -0,0 +1,309 @@
...
+				const to = Math.min(from + this.pageSize, maxRange);
+				const processedData = await this.fetchAndProcessRecords(from, to);
+				this.data = processedData;
+				this.currentPage = Math.ceil(from / this.pageSize) + 1
+				this.firstVal = from; // Set firstVal to searched value
+				this.lastVal = from + this.pageSize; // Calculate lastVal based on pageSize
+				this.displayRecords();
+				this.updatePageInfo();
+			}
+			else {
```
DevinFledermaus: move the `else` to the same line as the `}`

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		try {
+			this.adjustGridHeight();
+			await this.recordCount();
+			await this.fetchColumns();
+			await this.fetchRecords();
+			this.setupControls();
+		} catch (error) {
+			console.error('Error during initialization:', error);
+		}
+	}
```
DevinFledermaus: semicolons please

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	totalItems: number = 0;
+	columnNames: ColumnName[] = [];
+	maxGridHeight: number = 0;
+	firstVal: number = 0;
+	lastVal!: number;
+
+	constructor(pageSize: number) {
+		this.pageSize = pageSize;
+
+	}
```
DevinFledermaus: semicolons please

app.ts
```diff
@@ -0,0 +1,309 @@
...
+class ApiData {
+	// Properties to manage data and settings
+	pageSize: number;
+	currentPage: number = 1;
+	data: GridData[] = [];
+	totalItems: number = 0;
+	columnNames: ColumnName[] = [];
+	maxGridHeight: number = 0;
+	firstVal: number = 0;
+	lastVal!: number;
```
DevinFledermaus: that name might raise issues so remove the !

app.ts
```diff
@@ -0,0 +1,309 @@
...
+			this.currentPage = Math.floor(maxRange / this.pageSize) + 1; // Set currentPage to the last page
+			to = maxRange;
+		}
+		try {
+			const processedData = await this.fetchAndProcessRecords(from, to);
+			this.data = processedData;
+			this.displayRecords();
+			this.updatePageInfo();
+		} catch (error) {
+			alert('failed to fetch records ');
```
DevinFledermaus: You doing nothing with the error

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		}
+	}
+
+
```
DevinFledermaus: unnecessary space

app.ts
DevinFledermaus: If you set the variable, no need to provide the type

app.ts
```diff
@@ -0,0 +1,309 @@
...
+	//get records from API for fetch an search functionality 
+	async fetchAndProcessRecords(from: number = this.firstVal, to: number = this.lastVal) {
+		try {
+			$('#spinner').show()
+			$('#grid').hide()
+			const response = await this.fetchData(`http://localhost:2050/records?from=${from}&to=${to}`);
+			const res = JSON.parse(response);
+			const processedData = res.map((record: any) => {
+				const obj: GridData = {};
+				for (let j = 0; j < this.columnNames.length && j < record.length; j++) {
```
DevinFledermaus: Use a `for..of` loop instead

app.ts
```diff
@@ -0,0 +1,309 @@
...
+
+
+	async fetchRecords() {
+		const maxRange = this.totalItems - 1;
+		const from = this.firstVal;
+		let to = Math.min(from + this.pageSize, maxRange);
+		if (to >= maxRange) {
+			this.currentPage = Math.floor(maxRange / this.pageSize) + 1; // Set currentPage to the last page
+			to = maxRange;
+		}
```
DevinFledermaus: remember semicolons

app.ts
```diff
@@ -0,0 +1,309 @@
...
+		const pageCntrl = $('.grid-controls').innerHeight();
+		const screenHeight = $(window).innerHeight();
+		if (gridElement && pageCntrl !== undefined && screenHeight !== undefined) {
+			this.maxGridHeight = screenHeight - pageCntrl;
+			gridElement.style.height = `${this.maxGridHeight}px`;
+			gridElement.style.overflow = 'none';
+		}
+	}
+
+	private async fetchData(url: string): Promise<any> {
```
DevinFledermaus: we dont like making use of any too much, maybe give the types 
for example:
```suggestion
	private async fetchData(url: string): Promise<number> | Promise<string> {
```

app.ts
```diff
@@ -0,0 +1,309 @@
...
+
+
+	private displayRecords = () => {
+		const gridTemplate = new GridTemplate(this.columnNames, this.data);
+		gridTemplate.displayRecords();
+		this.updatePageInfo();
+	};
+
+	// Update the page information and records display based on the current state of the grid.
+	updatePageInfo() {
```
DevinFledermaus: should be above the private functions, check the style guide for reference

app.ts
```diff
@@ -0,0 +1,309 @@
...
+			const res = JSON.parse(response);
+			this.columnNames = res.map((columnName: any) => ({ name: columnName }));
+			this.data = new Array<GridData>(this.columnNames.length);
+		} catch (error) {
+			console.error('Failed to fetch columns', error);
+			throw error;
+		}
+	}
+	//get records from API for fetch an search functionality 
+	async fetchAndProcessRecords(from: number = this.firstVal, to: number = this.lastVal) {
```
DevinFledermaus: return types of the functions

DevinFledermaus (changes requested):
Try moving your two classes, ApiData and GridTemplate to seperate files

ApiData.ts
```diff
@@ -0,0 +1,201 @@
+class ApiData {
+	// Properties to manage data and settings
+	pageSize: number;
+	currentPage = 1;
```
FritzOnFire: Please always add types to class attributes, even thou you don't need them.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	data: GridData[] = [];
+	totalItems = 0;
```
FritzOnFire: Please always add types to class attributes, even thou you don't need them.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	columnNames: ColumnName[] = [];
+	maxGridHeight = 0;
```
FritzOnFire: Please always add types to class attributes, even thou you don't need them.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	firstVal = 0;
```
FritzOnFire: Please always add types to class attributes, even thou you don't need them.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	lastVal: number | undefined;
```
FritzOnFire: Nothing wrong with what you have here. I would personally initialize the value to `-1` and rather check for that. Checking for undefined on a number variable is not a lot of fun.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+
+	constructor(pageSize: number) {
+		this.pageSize = pageSize;
+	};
```
FritzOnFire: Please remove this one `;`.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	// Initialize method to set up the grid
+	async initialize(): Promise<void> {
```
FritzOnFire: Please add an empty line between these two functions.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	totalItems = 0;
+	columnNames: ColumnName[] = [];
+	maxGridHeight = 0;
+	firstVal = 0;
+	lastVal: number | undefined;
+
+	constructor(pageSize: number) {
+		this.pageSize = pageSize;
+	};
+	// Initialize method to set up the grid
```
FritzOnFire: Please change this to the javadoc format. This is unfortunately not in our style guide :( But it really should be.
```suggestion
	/** Initialize method to set up the grid */
```

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	async initialize(): Promise<void> {
+		try {
```
FritzOnFire: Instead of using `try-catch`, you should rather add `.catch` to your functions that return promises. But based on the way this is currently written, I would just remove the `try-catch` here, and add a `.catch` to your `apidata.initialize();` in app.ts and place your `console.error` in there.

Also, the user can not see this error, please show this error to the user after logging it to the console. You can just use alert (unless you want to do something fancy, but that is outside the scope of this ticket)

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			await this.fetchColumns();
+			await this.fetchRecords();
+			this.setupControls();
+		} catch (error) {
+			console.error('Error during initialization:', error);
+		}
+	}
+	// Method to fetch total record count from the server
+	async recordCount(): Promise<void> {
+		try {
```
FritzOnFire: Please remove the `try-catch` and rather add a `.catch` to `this.fetchData` with the contents of your current `catch`.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			const response = await this.fetchData('http://localhost:2050/recordCount');
+			this.totalItems = typeof response === 'number' ? response : parseInt(response as string, 10);
```
FritzOnFire: Please remove the `await` and the `const response`. Rather just return `this.fetchData` and add a `.then` to the end of your `this.fetchData`, that does what your function would have done. The you can also remove the `async` in front of your function declaration.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+		} catch (error) {
+			throw new Error('Failed to fetch record count.');
```
FritzOnFire:
```suggestion
			throw 'Failed to fetch record count.';
```

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			await this.recordCount();
+			await this.fetchColumns();
+			await this.fetchRecords();
+			this.setupControls();
+		} catch (error) {
+			console.error('Error during initialization:', error);
+		}
+	}
+	// Method to fetch total record count from the server
+	async recordCount(): Promise<void> {
```
FritzOnFire: Please add an empty line between these two functions.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			this.adjustGridHeight();
+			await this.recordCount();
+			await this.fetchColumns();
+			await this.fetchRecords();
+			this.setupControls();
+		} catch (error) {
+			console.error('Error during initialization:', error);
+		}
+	}
+	// Method to fetch total record count from the server
```
FritzOnFire: Please change this comment to the javadoc format

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	async recordCount(): Promise<void> {
+		try {
+			const response = await this.fetchData('http://localhost:2050/recordCount');
+			this.totalItems = typeof response === 'number' ? response : parseInt(response as string, 10);
+		} catch (error) {
+			throw new Error('Failed to fetch record count.');
+		}
+	}
+	//fectch column names
+	async fetchColumns(): Promise<void> {
```
FritzOnFire: Please add an empty line between these two functions.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	// Method to fetch total record count from the server
+	async recordCount(): Promise<void> {
+		try {
+			const response = await this.fetchData('http://localhost:2050/recordCount');
+			this.totalItems = typeof response === 'number' ? response : parseInt(response as string, 10);
+		} catch (error) {
+			throw new Error('Failed to fetch record count.');
+		}
+	}
+	//fectch column names
```
FritzOnFire: Please change the comment to the javadoc format

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	async fetchColumns(): Promise<void> {
+		try {
+			const response = await this.fetchData('http://localhost:2050/columns');
+			const res = JSON.parse(response as string);
```
FritzOnFire: Please use the `<>` way of type casting. https://github.com/IMQS/www-lib/tree/master/_codestyle#type-casting

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+		try {
+			const response = await this.fetchData('http://localhost:2050/recordCount');
+			this.totalItems = typeof response === 'number' ? response : parseInt(response as string, 10);
+		} catch (error) {
+			throw new Error('Failed to fetch record count.');
+		}
+	}
+	//fectch column names
+	async fetchColumns(): Promise<void> {
+		try {
```
FritzOnFire: Please remove the `try-catch` and rather add a `.catch` to your `this.fetchData` containing the same content as the catch that you had.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			const response = await this.fetchData('http://localhost:2050/columns');
+			const res = JSON.parse(response as string);
+			this.columnNames = res.map((columnName: any) => ({ name: columnName }));
+			this.data = new Array<GridData>(this.columnNames.length);
```
FritzOnFire: Please remove the `const response` and `await`, and rather return the `this.fetchData` and add a `.then` to the `this.fetchData` with the content of what the rest of the function was doing. Then you can also remove the `async` in front of your function declaration.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+		} catch (error) {
+			throw new Error('Failed to fetch columns.');
+		}
+	};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	//get records from API for fetch an search functionality 
+	async fetchAndProcessRecords(from: number, to: number): Promise<GridData[]> {
+		try {
+			$('#spinner').show()
```
FritzOnFire: Missing `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+				const obj: GridData = {};
+				let columnIndex = 0;
+				for (const column of this.columnNames) {
+					if (columnIndex < record.length) {
+						const columnName = column.name;
+						const columnValue = record[columnIndex];
+						obj[columnName] = columnValue;
+					}
+					columnIndex++;
+				}
```
FritzOnFire: If you need the index of the array you are looping over, then you should rather use a c-style loop, than creating a `columnIndex` variable.

Don't create the `columnName` and `columnValue` variables, if you are only going to use them once. Rather just use them directly when setting the values on obj.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	//fetch records from api
+	async fetchRecords(): Promise<void> {
+		const maxRange = this.totalItems - 1;
+		const from = this.firstVal;
+		let to = Math.min(from + this.pageSize, maxRange);
+		if (to >= maxRange) {
+			// Set currentPage to the last page
+			this.currentPage = Math.floor(maxRange / this.pageSize) + 1;
+			to = maxRange;
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+		try {
+			const processedData = await this.fetchAndProcessRecords(from, to);
+			this.data = processedData;
+			this.displayRecords();
+			this.updatePageInfo();
+		} catch (error) {
+			throw new Error('Failed to fetch records')
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+				// Calculate lastVal based on pageSize
+				this.lastVal = from + this.pageSize;
+				this.displayRecords();
+				this.updatePageInfo();
+			} else {
+				alert('Please enter values in the range (0-999999)');
+			}
+		} catch (error) {
+			throw new Error('Failed to search value');
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			this.currentPage = Math.floor(maxRange / this.pageSize) + 1;
+			to = maxRange;
+		};
+		try {
+			const processedData = await this.fetchAndProcessRecords(from, to);
+			this.data = processedData;
+			this.displayRecords();
+			this.updatePageInfo();
+		} catch (error) {
+			throw new Error('Failed to fetch records')
```
FritzOnFire: Missing `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	async searchRecords(searchValue: number): Promise<void> {
+		try {
+			// Maximum allowed Value
+			const maxRange = this.totalItems - 1;
+			if (searchValue >= 0 && searchValue <= maxRange) {
+				const from = searchValue;
+				const to = Math.min(from + this.pageSize, maxRange);
+				const processedData = await this.fetchAndProcessRecords(from, to);
+				this.data = processedData;
+				this.currentPage = Math.ceil(from / this.pageSize) + 1
```
FritzOnFire: Missing `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	}
+	//change grid height according to screen size
+	private adjustGridHeight(): void {
+		const gridElement = document.getElementById('grid');
+		const pageCntrl = $('.grid-controls').innerHeight();
+		const screenHeight = $(window).innerHeight();
+		if (gridElement && pageCntrl !== undefined && screenHeight !== undefined) {
+			this.maxGridHeight = screenHeight - pageCntrl;
+			gridElement.style.height = `${this.maxGridHeight}px`;
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			$('#overlay').show();
+			const response = await $.ajax({
+				url,
+				method: 'GET',
+			});
+			$('#overlay').hide();
+			return response;
+		} catch (error) {
+			throw error;
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+	}
+	private setupControls(): void {
+		$('#prevBtn').on('click', () => this.handlePageChange(-1));
+		$('#nextBtn').on('click', () => this.handlePageChange(1));
+		$(window).on('resize', debounce(this.handleResize, 350));
```
FritzOnFire: A debounce should ideally be around 50 ms, the lower the better, with an absolute max of 250 ms. Anything above that is not ok.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+		const newFirstVal = this.firstVal + delta * this.pageSize;
+		// If moving forward and newFirstVal exceeds total items, go to the first page
+		if (delta > 0 && newFirstVal > this.totalItems - 1) {
+			this.firstVal = 0;
+		} else if (delta < 0 && newFirstVal < 0) {
+			// If moving backward and newFirstVal goes below 0, go to the last page
+			this.firstVal = Math.max(0, this.totalItems - this.pageSize);
+		} else {
+			this.firstVal = Math.max(0, Math.min(newFirstVal, this.totalItems - 1));
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+		this.lastVal = this.firstVal + this.pageSize - 1;
+		this.currentPage = Math.floor(this.firstVal / this.pageSize) + 1;
+		this.fetchRecords();
+		this.updatePageInfo();
+	}
+	//for resizing the page 
+	private handleResize = (): void => {
```
FritzOnFire: Please don't make variables on a class that are just functions if you have no intention of changing the value, or allowing someone else to change the value.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			this.firstVal = newFirstValueIndex;
+			this.lastVal = newFirstValueIndex + newPageSize - 1;
+			// Fetch records, update page info, and adjust grid height
+			this.fetchRecords();
+			this.updatePageInfo();
+			this.adjustGridHeight();
+		};
+	}
+
+	private displayRecords = (): void => {
```
FritzOnFire: Please don't make variables on a class that are just functions if you have no intention of changing the value, or allowing someone else to change the value.

ApiData.ts
```diff
@@ -0,0 +1,201 @@
...
+			}
+			// Update firstVal, lastVal, and page size
+			this.pageSize = newPageSize;
+			this.firstVal = newFirstValueIndex;
+			this.lastVal = newFirstValueIndex + newPageSize - 1;
+			// Fetch records, update page info, and adjust grid height
+			this.fetchRecords();
+			this.updatePageInfo();
+			this.adjustGridHeight();
+		};
```
FritzOnFire: Please remove this one `;`

ApiData.ts
FritzOnFire, line 1: Most of the functions have a similar comment, so I stopped after a few :P. But I need you to
* add an empty line between functions.
* Change the comment of the function, if it has one, to the javadoc format.
* Remove any and all `try-catch` statements and replace them with `.then` and `.catch` and return the promise to remove the `async` in front of the function declaration. (I think `searchRecords` might be the only one that will still need `async` thou)

GridTemp.ts
```diff
@@ -0,0 +1,41 @@
+// Class to manage the grid template and display records
```
FritzOnFire: Please change this function to the javadoc format

GridTemp.ts
```diff
@@ -0,0 +1,41 @@
...
+class GridTemplate {
+	private columnNames: ColumnName[] = [];
+	private dataRecords: GridData[] = [];
+	// Initializes the column names and data records that will be used to display records in the grid.
+	constructor(columnNames: ColumnName[], dataRecords: GridData[]) {
```
FritzOnFire: Please add an empty line between your constructor and your variables.

GridTemp.ts
```diff
@@ -0,0 +1,41 @@
...
+		this.columnNames = columnNames;
+		this.dataRecords = dataRecords;
+	};
```
FritzOnFire: Please remove this one `;`

GridTemp.ts
```diff
@@ -0,0 +1,41 @@
...
+	// Display records in a grid in table format 
+	displayRecords(): void {
```
FritzOnFire: Please add an empty line between these two functions

GridTemp.ts
```diff
@@ -0,0 +1,41 @@
...
+					td.textContent = row[column.name];
+					tr.appendChild(td);
+				}
+				tbody.appendChild(tr);
+			}
+			table.appendChild(tbody);
+			// Append the table to the grid element
+			gridElement.appendChild(table);
+		}
+	};
```
FritzOnFire: Please remove the `;`

Styles.css
```diff
@@ -0,0 +1,86 @@
...
+
+tr {
+	border: 1px solid #333333;
+}
+
+thead {
+	background-color: #333333;
+	font-weight: 450;
+	color: white;
+
```
FritzOnFire: Please remove this one empty line

app.ts
```diff
@@ -0,0 +1,61 @@
...
+	return (...args: Parameters<F>): Promise<ReturnType<F>> => {
+		clearTimeout(timeout);
+
+		return new Promise((resolve) => {
+			timeout = setTimeout(() => {
+				resolve(func(...args));
+			}, waitFor);
+		});
+	};
+};
```
FritzOnFire: Please remove this one `;`

app.ts
```diff
@@ -0,0 +1,61 @@
+// Interface to define the structure of grid data
```
FritzOnFire: Please change this comment to the javadoc format

app.ts
```diff
@@ -0,0 +1,61 @@
...
+interface ColumnName {
+	name: string;
+}
+// Interface to define column names
```
FritzOnFire: Please change this comment to the javadoc format

app.ts
```diff
@@ -0,0 +1,61 @@
...
+interface GridData {
+	[key: string]: any;
+}
+// Debounce utility function to limit function execution frequency
```
FritzOnFire: Please change this comment to the javadoc format

app.ts
```diff
@@ -0,0 +1,61 @@
+// Interface to define the structure of grid data
+interface ColumnName {
+	name: string;
+}
+// Interface to define column names
+interface GridData {
```
FritzOnFire: Please add an empty line between these two interfaces.

app.ts
```diff
@@ -0,0 +1,61 @@
...
+	[key: string]: any;
+}
+// Debounce utility function to limit function execution frequency
+function debounce<F extends (...args: any) => any>(func: F, waitFor: number) {
```
FritzOnFire: Please add an empty line between this function and the interfaces.

app.ts
```diff
@@ -0,0 +1,61 @@
...
+		return new Promise((resolve) => {
+			timeout = setTimeout(() => {
+				resolve(func(...args));
+			}, waitFor);
+		});
+	};
+};
+// Constants for grid calculation
+const gridRatio = 9 / 20;// represents the ratio of the grid's height to the window's height. 
+const rowHeight = 16;
```
FritzOnFire: Please change these two variable names to all caps with underscores.

app.ts
```diff
@@ -0,0 +1,61 @@
...
+
+		return new Promise((resolve) => {
+			timeout = setTimeout(() => {
+				resolve(func(...args));
+			}, waitFor);
+		});
+	};
+};
+// Constants for grid calculation
+const gridRatio = 9 / 20;// represents the ratio of the grid's height to the window's height.
```
FritzOnFire: Please move the comment to above the variable and change the format to javadoc

app.ts
```diff
@@ -0,0 +1,61 @@
...
+		});
+	};
+};
+// Constants for grid calculation
+const gridRatio = 9 / 20;// represents the ratio of the grid's height to the window's height. 
+const rowHeight = 16;
+// Wait for the document to be ready
+$(document).ready(() => {
+	// Initialization and setup code
+	const windowHeight = Math.floor($(window).innerHeight() as number);
```
FritzOnFire: Please use `<>` type casting.

app.ts
```diff
@@ -0,0 +1,61 @@
...
+	const initialGridSize = Math.floor((windowHeight * gridRatio) / rowHeight);
+	const apidata = new ApiData(initialGridSize);
+	// Set up search button click handler
+	$('#searchBtn').on('click', () => {
+		const from = parseInt($('#fromInput').val() as string);
```
FritzOnFire: Please use `<>` type casting

app.ts
```diff
@@ -0,0 +1,61 @@
...
+			if (adjustedFrom + pageSize > maxRange) {
+				adjustedFrom = Math.max(0, maxRange - pageSize);
+				to = maxRange;
+			}
+			apidata.searchRecords(adjustedFrom);
+		} else if (from < 0 || from > maxRange) {
+			alert('please enter values in the range (0-999999)');
+			return;
+		} else if (isNaN(from)) {
+			alert('Please enter a numerical value ')
```

app.ts
```diff
@@ -0,0 +1,61 @@
...
+		} else {
+			console.error('error')
```

app.ts
```diff
@@ -0,0 +1,61 @@
...
+		};
```
FritzOnFire: Please remove this one `;`

index.html
```diff
@@ -1,13 +1,33 @@
 <!DOCTYPE html>
 <html>
+
 <head>
 	<title>JS Onboard Project</title>
 	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+	<link rel="stylesheet" href="/Styles.css">
+	<script src="app.js"></script>
```
FritzOnFire: `app.js` makes use of classes in `ApiData.js` and `GridTemp.js`. Please place it after those two.

And then you will have to move the interfaces and functions that are being mentioned by `ApiData.js` and `GridTemp.js` to either one of those two files, or a new file so that they can be available before the code in `ApiData.js` and `GridTemp.js` mentions them.

FritzOnFire (changes requested):
Full disclaimer. I have only made a lot of high level comments, I have not really looked at any logic. Will do that after these bulk changes are out of the way.

ApiData.ts
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	/** Initialize method to set up the grid */
+	async initialize(): Promise<void> {
+		Promise.resolve()
+			.then(() => this.adjustGridHeight())
+			.then(() => this.recordCount())
+			.then(() => this.fetchColumns())
+			.then(() => this.fetchRecords())
+			.then(() => this.setupControls())
```
anzelIMQS:
```suggestion
			.then(() => this.setupControls());
```
Missing semi-colon

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	constructor(pageSize: number) {
+		this.pageSize = pageSize;
+	}
+
+	/** Initialize method to set up the grid */
+	async initialize(): Promise<void> {
+		Promise.resolve()
+			.then(() => this.adjustGridHeight())
+			.then(() => this.recordCount())
+			.then(() => this.fetchColumns())
```
FritzOnFire: Please don't create a `Promise.resolve` just to chain with another promise :P. And `adjustGridHeight` does not return a promise, so no need to treat it as such. And please return the promise to allow other function to await on this function in case they need to.

```suggestion
		this.adjustGridHeight();
		return this.recordCount()
			.then(() => this.fetchColumns())
```

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	maxGridHeight: number = 0;
+	firstVal: number = 0;
+	lastVal: number = -1;
+
+	constructor(pageSize: number) {
+		this.pageSize = pageSize;
+	}
+
+	/** Initialize method to set up the grid */
+	async initialize(): Promise<void> {
```
FritzOnFire: No need for `async`, as you are not `await`ing or returning something that is not a promise.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+		Promise.resolve()
+			.then(() => this.adjustGridHeight())
+			.then(() => this.recordCount())
+			.then(() => this.fetchColumns())
+			.then(() => this.fetchRecords())
+			.then(() => this.setupControls())
```
FritzOnFire: Missing `;`

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	/**Method to fetch total record count from the server */
```
FritzOnFire: Needed a space, and no need to say that a method is a method
```suggestion
	/** Fetch total record count from the server */
```

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+			.then((response: any) => {
+				const totalItems = response;
+				this.totalItems = totalItems;
+			})
+			.catch(() => {
+				throw ('Failed to fetch record count.');
+			});
+	}
+
+	/**fectch column names*/
```
FritzOnFire: Please leave a space after the `/**` and before the `*/`

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+				const res = JSON.parse(response);
+				this.columnNames = res.map((columnName: any) => ({ name: columnName }));
+				this.data = new Array<GridData>(this.columnNames.length);
+			})
+			.catch(() => {
+				throw ('Failed to fetch columns.');
+			});
+	}
+
+	/**get records from API for fetch an search functionality*/
```
FritzOnFire: Please leave a space after the `/**` and before the `*/`

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+				$('#spinner').hide();
+				$('#grid').show();
+				return processedData;
+			})
+			.catch(() => {
+				throw ('Failed to fetch records');
+			});
+	}
+
+	/**fetch records from api*/
```
FritzOnFire: Please leave a space after the `/**` and before the `*/`

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+				this.data = processedData;
+				this.displayRecords();
+				this.updatePageInfo();
+			})
+			.catch(() => {
+				throw ('Failed to fetch records');
+			});
+	}
+
+	/**funtion to search through records using fromID*/
```
FritzOnFire: Please leave a space after the `/**` and before the `*/`. Also no need to call a function a function.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	fetchRecords(): Promise<void> {
+		const maxRange = this.totalItems - 1;
+		const from = this.firstVal;
+		let to = Math.min(from + this.pageSize, maxRange);
+
+		if (to >= maxRange) {
+			// Set currentPage to the last page
+			this.currentPage = Math.floor(maxRange / this.pageSize) + 1;
+			to = maxRange;
+		}
```
FritzOnFire: Should you not also update the value of `from`? And I guess `firstVal`? If not in this `if`, at least somewhere in the function. Otherwise you are not `fetching` enough records.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+					// Calculate lastVal based on pageSize
+					this.lastVal = from + this.pageSize - 1;
+					this.displayRecords();
+					this.updatePageInfo();
+				})
+				.catch(() => {
+					throw ('Failed to search value');
+				});
+		} else {
+			alert('Please enter values in the range (0-999999)');
```
FritzOnFire: Not all back-ends will have the same max id. You will have to change this to not be hard coded.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	/**funtion to search through records using fromID*/
+	searchRecords(searchValue: number): Promise<void> {
+		// Maximum allowed Value
+		const maxRange = this.totalItems - 1;
+
+		if (searchValue >= 0 && searchValue <= maxRange) {
+			const from = searchValue;
+			const to = Math.min(from + this.pageSize, maxRange);
+
+			return this.fetchAndProcessRecords(from, to)
```
FritzOnFire: Please adjust your `from` to allow fetching from the back-end as many records as you need to display. As currently this looks like it could only return 2 records when you might need 40 to fill up the grid.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+				.then((processedData) => {
```
FritzOnFire: Either remove the `()` around `processedData` or specify a type.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+		let to = Math.min(from + this.pageSize, maxRange);
+
+		if (to >= maxRange) {
+			// Set currentPage to the last page
+			this.currentPage = Math.floor(maxRange / this.pageSize) + 1;
+			to = maxRange;
+		}
+
+		return this.fetchAndProcessRecords(from, to)
+			.then((processedData) => {
```
FritzOnFire: Either remove the `()` around `processedData` or specify a type.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	private setupControls(): void {
+		$('#prevBtn').on('click', () => this.handlePageChange(-1));
+		$('#nextBtn').on('click', () => this.handlePageChange(1));
+		$(window).on('resize', debounce(this.handleResize, 100));
+	}
+
+	private handlePageChange(delta: number): void {
+
+		if (delta > 0 && this.firstVal + delta * this.pageSize > this.totalItems - 1) {
+			this.firstVal = 0;
```
FritzOnFire: Help me if I am wrong here... but

* if delta is positive, so next page.
* and the current `firstVal` plus the amount of records you need to show is more than the max
* You reset `firstVal` ALL they way back to zero...

This does not sound correct. Should you not be setting it to max - the amount of records you need to show?

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+		} else if (delta < 0 && this.firstVal + delta * this.pageSize < 0) {
+			this.firstVal = Math.max(0, this.totalItems - this.pageSize);
```
FritzOnFire: Help me if I am wrong here... but

* if delta is negative, so previous page.
* and the current `firstVal`  minus the amount of records you need to show is less than zero
* You set `firstVal` to the highest possible value it can be...

This does not sound correct. Should you not be setting it to zero?

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+		} else {
+			this.firstVal = Math.max(0, Math.min(this.firstVal + delta * this.pageSize, this.totalItems - 1));
```
FritzOnFire: Nice, very clean way of ensuring you have a valid value :)

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+		}
+
+		this.lastVal = this.firstVal + this.pageSize ;
```
FritzOnFire: Please remove the space before the `;`

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	private handleResize = (): void => {
+		const newGridSize = Math.floor((Math.floor(<number>($(window).innerHeight())) * GRID_RATIO) / ROW_HEIGHT);
+
+		// Check if the new grid size is non-negative
+		if (newGridSize >= 0) {
+			// Adjust firstVal for the last page
+			if (this.firstVal + newGridSize > this.totalItems) {
+				this.firstVal = Math.max(this.totalItems - newGridSize);
```
FritzOnFire: I don't think you need the `max` here

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+			this.firstVal = 0;
+		} else if (delta < 0 && this.firstVal + delta * this.pageSize < 0) {
+			this.firstVal = Math.max(0, this.totalItems - this.pageSize);
+		} else {
+			this.firstVal = Math.max(0, Math.min(this.firstVal + delta * this.pageSize, this.totalItems - 1));
+		}
+
+		this.lastVal = this.firstVal + this.pageSize ;
+		this.currentPage = Math.floor(this.firstVal / this.pageSize) + 1;
+		this.fetchRecords();
```
FritzOnFire: `fetchRecords` can `throw` an error, you need to catch it and handle it.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+		this.updatePageInfo();
```
FritzOnFire: I think it would be better if you only did `updatePageInfo` after `fetchRecords` has finished. Just in case it fails.

ApiData.ts
FritzOnFire: Do you not also need `-1` here?

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+				this.firstVal = Math.max(this.totalItems - newGridSize);
+			}
+
+			this.pageSize = newGridSize;
+			this.lastVal = this.firstVal + newGridSize - 1;
+
+			// Fetch records, update page info, and adjust grid height
+			this.fetchRecords();
+			this.updatePageInfo();
+			this.adjustGridHeight();
```
FritzOnFire, line 231: Should you not call `adjustGridHeight` before `fetchRecords` and `updatePageInfo`? As it does not require any of the previous to to function.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+			// Adjust firstVal for the last page
+			if (this.firstVal + newGridSize > this.totalItems) {
+				this.firstVal = Math.max(this.totalItems - newGridSize);
+			}
+
+			this.pageSize = newGridSize;
+			this.lastVal = this.firstVal + newGridSize - 1;
+
+			// Fetch records, update page info, and adjust grid height
+			this.fetchRecords();
```
FritzOnFire: `fetchRecords` can `throw` an error. You need to catch it and handle it.

ApiData.ts
```diff
@@ -0,0 +1,209 @@
...
+			this.updatePageInfo();
```
FritzOnFire: I think it would be better if you only did `updatePageInfo` after `fetchRecords` has finished. Just in case it fails.

GridTemp.ts
```diff
@@ -0,0 +1,44 @@
...
+	private columnNames: ColumnName[] = [];
+	private dataRecords: GridData[] = [];
+
+	/** Initializes the column names and data records that will be used to display records in the grid. */
+	constructor(columnNames: ColumnName[], dataRecords: GridData[]) {
+		this.columnNames = columnNames;
+		this.dataRecords = dataRecords;
+	}
+
+	/*Display records in a grid in table format*/
```
FritzOnFire: You are missing one `*` at the beginning of your javadoc comment, and please add spaces after the `/**` and before the `*/`

Interfaces.ts
```diff
@@ -0,0 +1,9 @@
+/**Interface to define the structure of grid data*/
```
FritzOnFire: Please leave a space after the `/**` and before the `*/`. Also, I think this comment belongs on a different `interface`

Interfaces.ts
```diff
@@ -0,0 +1,9 @@
...
+interface ColumnName {
+	name: string;
+}
+
+/**Interface to define column names*/
```
FritzOnFire: Please leave a space after the `/**` and before the `*/`. Also, I think this comment belongs on a different `interface`

app.ts
```diff
@@ -0,0 +1,61 @@
+/**  Debounce utility function to limit function execution frequency*/
```
FritzOnFire: One to many spaces at the start of this comment, and please add a space before the `*/`

app.ts
```diff
@@ -0,0 +1,61 @@
...
+
+		return new Promise((resolve) => {
+			timeout = setTimeout(() => {
+				resolve(func(...args));
+			}, waitFor);
+		});
+	};
+}
+
+/**  Constants for grid calculation
```
FritzOnFire: Once space to many at the start of the comment

app.ts
```diff
@@ -0,0 +1,61 @@
...
+ * GRID_RATIO represents the ratio of the grid's height to the window's height.
+*/
```
FritzOnFire: Please indent this with a space.

app.ts
```diff
@@ -0,0 +1,61 @@
...
+const GRID_RATIO = 9/20 ;
```
FritzOnFire: Please remove the space before the `;`

app.ts
```diff
@@ -0,0 +1,61 @@
...
+const ROW_HEIGHT = 16;
+
+/**  Wait for the document to be ready*/
```
FritzOnFire: One space to many at the start of this comment, and please add a space before the `*/`

app.ts
```diff
@@ -0,0 +1,61 @@
...
+		if (!isNaN(from) && from >= 0 && from <= maxRange) {
+			let to = Math.min(from + pageSize, maxRange);
+			let adjustedFrom = from;
+			if (adjustedFrom + pageSize > maxRange) {
+				adjustedFrom = Math.max(0, maxRange - pageSize);
+				to = maxRange;
+			}
+			apidata.searchRecords(adjustedFrom);
+		} else if (from < 0 || from > maxRange) {
+			alert('please enter values in the range (0-999999)');
```
FritzOnFire: Not all back-ends will have the same max id. You will have to change this to not be hard coded.

app.ts
```diff
@@ -0,0 +1,61 @@
...
+	const initialGridSize = Math.floor((<number>($(window).innerHeight()) * GRID_RATIO) / ROW_HEIGHT) ;
+	const apidata = new ApiData(initialGridSize);
+	// Set up search button click handler
+	$('#searchBtn').on('click', () => {
+
+		const from = parseInt(<string>($('#fromInput').val()));
+		const pageSize = apidata.pageSize;
+		const maxRange = apidata.totalItems - 1;
+		if (!isNaN(from) && from >= 0 && from <= maxRange) {
+			let to = Math.min(from + pageSize, maxRange);
```
FritzOnFire: I'm not sure what the point of this `to` is?

index.html
```diff
@@ -1,13 +1,34 @@
 <!DOCTYPE html>
 <html>
+
 <head>
 	<title>JS Onboard Project</title>
 	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+	<link rel="stylesheet" href="/Styles.css">
+	<script src="/Interfaces.js"></script>
```
FritzOnFire: I think you can remove the `/`

index.html
```diff
@@ -1,13 +1,34 @@
...
+	<script src="ApiData.js"></script>
```
FritzOnFire, line 10: `ApiData` mentions `GRID_RATIO` which is defined in `app`. So you either have to

* move `ApiData` after `app` (I would not recommend this)
* or move `GRID_RATIO` to `ApiData`
* or create a different file to contain it and pull in that file before `ApiData`.

(The same is true for `ROW_HEIGHT` and `debounce`)

FritzOnFire (changes requested):

ApiData.ts
Ntobe99: Is it necessary to implement error handling here if  I have implemented it within the fetchRecords function?

ApiData.ts
```diff
@@ -0,0 +1,242 @@
...
+				return processedData;
+			})
+			.catch(() => {
+				throw ('Failed to fetch records');
+			});
+	}
+
+	/** fetch records from api */
+	fetchRecords(): Promise<void> {
+		const maxRange = this.totalItems - 1;
```
CelesteNaude: `maxRange` is being set in a lot of places to `maxRange = this.totalItems - 1`. I think it's worth it to set this value once (where you get the record count) as a class variable.

ApiData.ts
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+
+	private setupControls(): void {
+		$('#prevBtn').on('click', () => this.handlePageChange(-1));
+		$('#nextBtn').on('click', () => this.handlePageChange(1));
+		$(window).on('resize', debounce(this.handleResize, 100));
+	}
+
+	private handlePageChange(delta: number): void {
```
CelesteNaude, line 189: Please add some comments for your pagination.

app.ts
```diff
@@ -0,0 +1,38 @@
+/** Wait for the document to be ready */
+$(document).ready(() => {
+	// Initialization and setup code
+	const initialGridSize = Math.floor((<number>($(window).innerHeight()) * GRID_RATIO) / ROW_HEIGHT) ;
+	const apidata = new ApiData(initialGridSize);
+	// Set up search button click handler
+	$('#searchBtn').on('click', () => {
```
CelesteNaude, line 8: Search function can also be handled where the rest of the grid is being handled, would say it's part of grid functionality and should be included in its class.

CelesteNaude (changes requested):

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+	columnNames: ColumnName[] = [];
+	maxGridHeight: number = 0;
+	firstVal: number = 0;
+	lastVal: number = -1;
+	maxRange: number;
+
+	constructor(pageSize: number) {
+		this.pageSize = pageSize;
+		this.maxRange = 0;
+
```
FritzOnFire: Please remove this one empty line.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+	/** Initialize method to set up the grid */
+	initialize(): Promise<void> {
+		this.adjustGridHeight();
+		return this.recordCount()
+			.then(() => this.fetchColumns())
+			.then(() => this.fetchRecords())
+			.then(() => this.setupControls());
+	}
+	/** Fetch total record count from the server */
+	recordCount(): Promise<void> {
```
FritzOnFire: Please add an empty line between functions.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+		return this.fetchData('http://localhost:2050/recordCount')
+			.then((response: any) => {
```
FritzOnFire: You specified the return type for `fetchData` as either a `number` or `string`, you can't specify `any` here.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+			.catch(error => {
+				console.error('Failed to fetch record count:', error);
+				alert('Failed to fetch record count.');
+			});
+	}
+
+	/** fectch column names */
+	fetchColumns(): Promise<void> {
+		return this.fetchData('http://localhost:2050/columns')
+			.then((response: any) => {
```
FritzOnFire: You specified the return type for `fetchData` as either a `number` or `string`, you can't specify `any` here.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+				const res = JSON.parse(response);
+				this.columnNames = res.map((columnName: any) => ({ name: columnName }));
```
FritzOnFire: Pretty sure you know what the type of `columnName` is. Or, at least, what it should be for your code to work.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+			});
+	}
+
+	/** get records from API for fetch and search functionality */
+	fetchAndProcessRecords(from: number, to: number): Promise<GridData[]> {
+		$('#spinner').show();
+		$('#grid').hide();
+
+		return this.fetchData(`http://localhost:2050/records?from=${from}&to=${to}`)
+			.then((response: any) => {
```
FritzOnFire: You specified the return type for `fetchData` as either a `number` or `string`, you can't specify `any` here.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+				const totalItems = response;
+				this.totalItems = totalItems;
+			})
+			.catch(error => {
+				console.error('Failed to fetch record count:', error);
+				alert('Failed to fetch record count.');
+			});
+	}
+
+	/** fectch column names */
```
FritzOnFire: Please also mention the fact that you empty out and reinitialize the data for the table.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+
+	/** get records from API for fetch and search functionality */
+	fetchAndProcessRecords(from: number, to: number): Promise<GridData[]> {
+		$('#spinner').show();
+		$('#grid').hide();
+
+		return this.fetchData(`http://localhost:2050/records?from=${from}&to=${to}`)
+			.then((response: any) => {
+				const res = JSON.parse(response);
+				const processedData = res.map((record: any) => {
```
FritzOnFire: Pretty sure you know what the type of `record` is. Or at least, what it should be for your code to work.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+	/** fetch records from api */
+	fetchRecords(): Promise<void> {
+		this.maxRange = this.totalItems - 1;
+		let from = this.firstVal;
+		let to = Math.min(from + this.pageSize, this.maxRange);
+
+		if (to >= this.maxRange) {
+			const lastPage = Math.floor(this.maxRange / this.pageSize) + 1;
+			this.currentPage = lastPage;
+			from = Math.min(this.maxRange - this.pageSize);
```
FritzOnFire: Not sure why you are using `Math.min` here.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+				$('#grid').show();
+				return processedData;
+			})
+			.catch(() => {
+				throw ('Failed to fetch records');
+			});
+	}
+
+	/** fetch records from api */
+	fetchRecords(): Promise<void> {
```
FritzOnFire: This function name and comment does not really match up what the function is doing... Yes, it is fetching records, but it is doing that by calling `fetchAndProcessRecords` and then it is also displaying the records and updating the page info. Just feel like if someone read the function name and  comment, they would assume it is doing only that, and not also something else.

Please just add a bit more to the function name and comment.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+			to = this.maxRange;
+		}
+
+		return this.fetchAndProcessRecords(from, to)
+			.then(processedData => {
+				this.data = processedData;
+				this.displayRecords();
+				this.updatePageInfo();
+			})
+			.catch((error) => {
```
FritzOnFire: Either remove the `()` around `error` or specify the type.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+	recordCount(): Promise<void> {
+		return this.fetchData('http://localhost:2050/recordCount')
+			.then((response: any) => {
+				const totalItems = response;
+				this.totalItems = totalItems;
+			})
+			.catch(error => {
+				console.error('Failed to fetch record count:', error);
+				alert('Failed to fetch record count.');
+			});
```
FritzOnFire: So, by catching the error here, and more specifically, not throwing an error here, the calling function has no idea something went wrong, and will try and continue executing the rest of your code. This is usually a good thing, as we don't want a front-end app to break just because one part of it is broken.

But when I look at `initialize`, it eventually calls `fetchRecords`, and `fetchRecords` definitely does need `totalItems` to have a value. I kinda feel that your program should not try and call `fetchRecords` if `recordCount` failed. And for that you are going to have to throw an error here.

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+				.then((processedData) => {
+					this.data = processedData;
+					this.currentPage = Math.ceil(from / this.pageSize) + 1;
+					// Set firstVal to searched value
+					this.firstVal = from;
+					// Calculate lastVal based on pageSize
+					this.lastVal = from + this.pageSize - 1;
+					this.displayRecords();
+					this.updatePageInfo();
+					//empty search input after searching
```
FritzOnFire:
```suggestion
					// empty search input after searching 
```

ApiData.ts
```diff
@@ -0,0 +1,266 @@
...
+			// If delta is negative then reset firstVal to 0 and disable prev button 
+			this.firstVal = 0;
+			prevBtn.attr("disabled", "disabled");
+			nextBtn.attr("disabled", null);
+		} else {
+
+			this.firstVal = Math.max(0, Math.min(this.firstVal + delta * this.pageSize, this.maxRange));
+			prevBtn.attr("disabled", null);
+			nextBtn.attr("disabled", null);
+
```
FritzOnFire: Please remove this one empty line

app.ts
```diff
@@ -0,0 +1,20 @@
...
+		let from = parseInt(<string>($('#fromInput').val()));
+		apidata.searchRecords(from);
+	});
+	// Initialize the grid
+	apidata.initialize()
+		.catch((error) => {
+			console.error('Error during initialization:', error);
+			alert(error);
+		});
+	//overlay when the page is still getting ready
```
```suggestion
	// overlay when the page is still getting ready
```

FritzOnFire (changes requested):

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+		const from = this.firstVal;
+		let to = Math.min(from + this.pageSize, this.maxRange);
+		$('#pageInfo').text(`${pageInfo}`);
+		$('.records').text(`Showing records ${from} to ${to}`);
+	}
+
+	private setupControls(): void {
+		$('#prevBtn').on('click', () => this.handlePageChange(-1));
+		$('#nextBtn').on('click', () => this.handlePageChange(1));
+		$(window).on('resize', debounce(this.handleResize.bind(this), 100));
```
FritzOnFire: We prefer to use arrow functions instead of using `.bind(this)`.
```suggestion
		$(window).on('resize', debounce(() => { this.handleResize(); }, 100));
```

FritzOnFire (approved):
One small comment, but from my side I think you are done.

@anzelIMQS @CelesteNaude Can I ask both of you to give this a once over, just in case there is anything major that I looked over. But I feel that we have covered enough to call this done.

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+				alert('Failed to fetch record count.');
+				throw error;
+			});
+	}
+
+	/** Use the fetchData() func to make an HTTP request to the API endpoint and process the data */
+	fetchColumns(): Promise<void> {
+		return this.fetchData('http://localhost:2050/columns')
+			.then((response: number | string) => {
+				const res = JSON.parse(<string>(response));
```
anzelIMQS:
```suggestion
			.then((response: string) => {
				const res = JSON.parse(response);
```
Isn't just always a string? I would think that the `/columns` response brings back a number.

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+			.then(() => this.fetchAndDisplayRecords())
+			.then(() => this.setupControls());
+	}
+
+	/** Fetch total record count from the server,fetches data from an API and populates class properties */
+	recordCount(): Promise<void> {
+		return this.fetchData('http://localhost:2050/recordCount')
+			.then((response: number | string) => {
+				const totalItems = <number>response;
+				this.totalItems = totalItems;
```
anzelIMQS:
```suggestion
			.then((response: number) => {
				this.totalItems = response;
```
I also think this can only be a number once it is successful.

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	/** Get records from API for fetch and search functionality */
+	fetchAndProcessRecords(from: number, to: number): Promise<GridData[]> {
+		$('#spinner').show();
+		$('#grid').hide();
+
+		return this.fetchData(`http://localhost:2050/records?from=${from}&to=${to}`)
+			.then((response: number | string) => {
+				const res = JSON.parse(<string>(response));
```
anzelIMQS:
```suggestion
			.then((response: string) => {
				const res = JSON.parse(response);
```
It will never return a number and if one day it does, we know something fishy is going on and at least you would log that?

I would suggest `return response.json()` instead of `JSON.parse(response)` and handling your promises' error in your catch. Just a suggestion though. You don't have to as long as you log the error in your catch..

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+						obj[this.columnNames[j].name] = record[j];
+					}
+					return obj;
+				});
+				$('#spinner').hide();
+				$('#grid').show();
+				return processedData;
+			})
+			.catch(() => {
+				throw ('Failed to fetch records');
```
anzelIMQS:
```suggestion
			.catch(error => {
				throw ('Failed to fetch records: ', error);
```
I mean, wouldn't you like to know the error? We know if failed but why?

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+		}
+
+		return this.fetchAndProcessRecords(from, to)
+			.then(processedData => {
+				this.data = processedData;
+				this.displayRecords();
+				this.updatePageInfo();
+			})
+			.catch(error => {
+				console.error('Failed to fetch records:', error);
```
anzelIMQS, line 120: Technically speaking, what do you think your `console.error` will look like since you have nested promises?
Think about it 🤔

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+			this.currentPage = lastPage;
+			from = this.maxRange - this.pageSize;
+			to = this.maxRange;
+		}
+
+		return this.fetchAndProcessRecords(from, to)
+			.then(processedData => {
+				this.data = processedData;
+				this.displayRecords();
+				this.updatePageInfo();
```
anzelIMQS: So `this.displayRecords()` calls `this.updatePageInfo() `and then we call it again?

Sounds like double work to me 🥵

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+			return this.fetchAndProcessRecords(from, from + pageSize)
+				.then(processedData => {
+					this.data = processedData;
+					this.currentPage = Math.ceil(from / this.pageSize) + 1;
+					// Set firstVal to searched value
+					this.firstVal = from;
+					// Calculate lastVal based on pageSize
+					this.lastVal = from + this.pageSize - 1;
+					this.displayRecords();
+					this.updatePageInfo();
```
anzelIMQS: Same vibe as previously mentioned

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+				})
+				.catch(error => {
+					console.error("Error fetching records while resizing:", error);
+					alert('Error occured while resizing!');
+				});
+		}
+	}
+
+	private displayRecords(): void {
+		const gridTemplate = new GridTemplate(this.columnNames, this.data);
```
anzelIMQS, line 245: So we are creating the GridTemplate class everytime we displayRecords()?

I would have loved that we could just set the dataRecords every time we update the display because `columnNames` never changes throughout this life. It just sounds like a lot of objects being created and created and created...

ApiData.ts
```diff
@@ -0,0 +1,264 @@
...
+			prevBtn.attr("disabled", null);
+			nextBtn.attr("disabled", null);
+		}
+
+		this.lastVal = this.firstVal + delta * this.pageSize;
+		this.currentPage = Math.floor(this.firstVal / this.pageSize) + 1;
+
+		this.fetchAndDisplayRecords()
+			.then(() => {
+				this.updatePageInfo();
```
anzelIMQS: Please look at the number of times you call `this.updatePageInfo()` and see if it really is necessary? I would think fetchAndDisplayRecords() already called it for you and `this.displayRecords()` as well and now `this.handlePageChange()` as well.

anzelIMQS (changes requested):

CelesteNaude (approved):
Couldn't find any further major changes.

ApiData.ts
```diff
@@ -0,0 +1,269 @@
...
+			.then(() => this.fetchAndDisplayRecords())
+			.then(() => this.setupControls());
+	}
+
+	/** Fetch total record count from the server,fetches data from an API and populates class properties */
+	recordCount(): Promise<void> {
+		return this.fetchNumData('http://localhost:2050/recordCount')
+			.then((response: number) => {
+				const totalItems =response;
+				this.totalItems = totalItems;
```
anzelIMQS:
```suggestion
				this.totalItems = response;
```

ApiData.ts
```diff
@@ -0,0 +1,269 @@
...
+				console.error('Failed to fetch record count:', error);
+				alert('Failed to fetch record count.');
+				throw error;
+			});
+	}
+
+	/** Use the fetchData() func to make an HTTP request to the API endpoint and process the data */
+	fetchColumns(): Promise<void> {
+		return this.fetchStrData('http://localhost:2050/columns')
+			.then((response:string) => {
```
anzelIMQS:
```suggestion
			.then((response: string) => {
```

ApiData.ts
```diff
@@ -0,0 +1,269 @@
...
+			});
+	}
+
+	/** Get records from API for fetch and search functionality */
+	fetchAndProcessRecords(from: number, to: number): Promise<GridData[]> {
+		$('#spinner').show();
+		$('#grid').hide();
+
+		return this.fetchStrData(`http://localhost:2050/records?from=${from}&to=${to}`)
+			.then((response:string) => {
```
anzelIMQS:
```suggestion
			.then((response: string) => {
```

ApiData.ts
```diff
@@ -0,0 +1,269 @@
...
+		// Maximum allowed Value
+		const maxRange = this.maxRange;
+
+		if (searchValue >= 0 && searchValue <= maxRange) {
+			let from = searchValue;
+			const pageSize = this.pageSize;
+
+			if (from + pageSize > maxRange) {
+				from = Math.max(0, maxRange - pageSize);
+			}
```
anzelIMQS: So how much different is this logic compared to your `fetchAndDisplayRecords()` logic?

The way I see it you can set `this.firstVal = searchValue` and let nature take its course by just calling  `fetchAndDisplayRecords()`, or am I missing something?

Similar to what you did with `handleResize`

ApiData.ts
```diff
@@ -0,0 +1,269 @@
...
+
+			return this.fetchAndProcessRecords(from, from + pageSize)
+				.then(processedData => {
+					this.data = processedData;
+					this.currentPage = Math.ceil(from / this.pageSize) + 1;
+					// Set firstVal to searched value
+					this.firstVal = from;
+					// Calculate lastVal based on pageSize
+					this.lastVal = from + this.pageSize - 1;
```
anzelIMQS: With the above comment being said. This could then be executed before the time and life would be good, right?

anzelIMQS (changes requested):
Something small but I think after this you should be home free.

ApiData.ts
```diff
@@ -0,0 +1,251 @@
...
+			})
+			.catch(error => {
+				console.error('Failed to fetch records: ', error);
+				throw new Error('Failed to fetch records: ' + error);
+			});
+	}
+
+	/** Fetches records using fetchAndProcessRecords(), processes them, displays them, and updates page information. */
+	fetchAndDisplayRecords(): Promise<void> {
+		this.maxRange = this.totalItems - 1;
```
anzelIMQS: Please move this line to your `recordCount()` as this value does not change but you are setting it over and over again with the same value when you do a `fetchAndDisplayRecords`.

anzelIMQS (approved):
LGTM, just the one comment but no other major issues to comment on.

ApiData.ts
```diff
@@ -0,0 +1,250 @@
...
+	recordCount(): Promise<void> {
+		return this.fetchNumData('http://localhost:2050/recordCount')
+			.then((response: number) => {
+				this.totalItems = response;
+				this.maxRange = this.totalItems - 1;
+			})
+			.catch(error => {
+				console.error('Failed to fetch record count:', error);
+				alert('Failed to fetch record count.');
+				throw error;
```
FritzOnFire: ???

You handle the error... but then you throw it anyway? Rather just throw an error here, and handle the error as late as possible.

ApiData.ts
```diff
@@ -0,0 +1,250 @@
...
+
+	/** Get records from API for fetch and search functionality */
+	fetchAndProcessRecords(from: number, to: number): Promise<GridData[]> {
+		$('#spinner').show();
+		$('#grid').hide();
+
+		return this.fetchStrData(`http://localhost:2050/records?from=${from}&to=${to}`)
+			.then((response: string) => {
+				const res = JSON.parse(response);
+				const processedData = res.map((record: string) => {
```
FritzOnFire: I don't think the type you have for `record` is correct

ApiData.ts
```diff
@@ -0,0 +1,250 @@
...
+			this.firstVal = searchValue;
+			if (searchValue + this.pageSize > this.maxRange) {
+				this.firstVal = Math.max(0, this.maxRange - this.pageSize + 1);
+			}
+			this.currentPage = Math.ceil(this.firstVal / this.pageSize) + 1;
+			// empty search input after searching 
+			$('#fromInput').val('');
+			return this.fetchAndDisplayRecords();
+		} else {
+			alert(`Error while searching, please enter values in the range (0-${this.maxRange})`);
```
FritzOnFire, line 137: You don't need to make an changes, just a comment.

I find it interesting that you would handle the error here instead of throwing it and letting the calling function deal with it, but it kinda makes sens, I can't imagine doing anything after the function that needs it to succeed.

ApiData.ts
```diff
@@ -0,0 +1,250 @@
...
+		return response;
+	}
+
+	private async fetchNumData(url: string): Promise<number> {
+		const response = await $.ajax({
+			url,
+			method: 'GET',
+		});
+		return response;
+	}
```
FritzOnFire, line 159: I really want to comment on how you don't know if `response` is actually a number... but I am finding it hard to phrase it... but I guess for this project it is fine.

FritzOnFire (changes requested):
There are some comments, but I don't think they are that major.

I would once again say that, you are done.

PR #47, opened 2023-08-11
Updated Project

ThokozaniNqwili:


app.ts
```diff
@@ -0,0 +1,199 @@
+import { ajax, css } from "jquery";
+var firstNumber: number = 0;
+var lastNumber: number;
+var calculatedRows: number | null = null;
+var count: number | null = null;
```
DevinFledermaus: try to stick to two global variables. And make use of `let` instead of `var`

app.ts
```diff
@@ -0,0 +1,199 @@
...
+	let rowHeight = 35
+	if (availableHeight <= 0) {
+		return 0;
+	} else {
+		let maxRows = Math.floor(availableHeight / rowHeight);
+		return maxRows; // returns the maximum rows that can fit into screen
+	}
+}
+
+let resizeTimeout: number;
```
DevinFledermaus: Another global variable

app.ts
```diff
@@ -0,0 +1,199 @@
...
+
+async function displayRecords(): Promise<void> { //displays records from firstNumber to lastNumber
+	try {
+		$('#loader').show()
+		count = await fetchRecordCount() - 1;
+		calculatedRows = await adjustRowsByScreenHeight();
+		const inputValue = $('#searchInput').val() as number;
+		if (calculatedRows === 0) {
+			lastNumber = firstNumber
+
```
DevinFledermaus, line 70: remove space

app.ts
```diff
@@ -0,0 +1,199 @@
...
+			$('#page').append(`Showing record: ${firstNumber} - ${lastNumber}`); // changes the record range showing
+			$('#loader').hide()
+		} else {
+			firstNumber = count - (calculatedRows - 1)
+			lastNumber = count;
+			records = await fetchRecords(firstNumber, count);
+			$('#page').empty();
+			$('#page').append(`Showing record: ${firstNumber} - ${count}`);
+			$('#loader').hide()
+
```
DevinFledermaus: remove space

app.ts
```diff
@@ -0,0 +1,199 @@
...
+		}
+
+
```
DevinFledermaus, line 123: no need for double spacing

app.ts
```diff
@@ -0,0 +1,199 @@
...
+		tbody.empty();
+		for (let r = 0; r < records.length; r++) {
+			$("tbody").append(`<tr class="row"></tr>`); // creates row for each record
+			const lastRow = $(".row:last");
+			for (let i = 0; i < records[r].length; i++) {
+				lastRow.append(`<td>${records[r][i]}</td>`); //assign each record to their column in a specified row
+			}
+			if (records[r].includes(inputValue)) {
+
+				lastRow.css('background-color', '#DDC0B4'); // hightlights the searched row
```
DevinFledermaus: remove spaces around

app.ts
```diff
@@ -0,0 +1,199 @@
...
+			const colArray = columns; // assigns the columns to an array
+
+			for (let c = 0; c < colArray.length; c++) {
+				$(".head").append(`<th>${colArray[c]}</th>`); // creates a single column for each heading
+			}
+		})
+}
+fetchColumns()
+
+async function adjustRowsByScreenHeight(): Promise<number> { // calculates the number of rows that can fit the screen
```
DevinFledermaus: Why is this an async?

app.ts
```diff
@@ -0,0 +1,199 @@
...
+			lastNumber = firstNumber
+
+		}
+		if (!lastNumber) {
+			lastNumber = firstNumber + (calculatedRows - 1);
+		}
+		if (firstNumber < 0) {
+			firstNumber = 0;
+			lastNumber = firstNumber + (calculatedRows - 1);
+		}
```
DevinFledermaus: move the `else` on the same line as the `}` please

app.ts
```diff
@@ -0,0 +1,199 @@
...
+	try {
+		$('#loader').show()
+		count = await fetchRecordCount() - 1;
+		calculatedRows = await adjustRowsByScreenHeight();
+		const inputValue = $('#searchInput').val() as number;
+		if (calculatedRows === 0) {
+			lastNumber = firstNumber
+
+		}
+		if (!lastNumber) {
```
DevinFledermaus: this and the `else` does exactly the same

app.ts
```diff
@@ -0,0 +1,199 @@
+import { ajax, css } from "jquery";
+var firstNumber: number = 0;
+var lastNumber: number;
```
DevinFledermaus: would be nice to have a default

app.ts
```diff
@@ -0,0 +1,199 @@
...
+	await displayRecords();
+}
+$('#closeModalBtn').on("click", () => {
+	$('.modal').css('display', 'none') // closes modal
+});
+$('.btnSearch').on('click', async (event: any) => {
+	event.preventDefault();
+	let inputValue = $('#searchInput').val() as number;
+	await updateRecordsAndResize(inputValue); // calls to calculate the range once button is clicked
+
```
DevinFledermaus, line 154: remove space

index.html
```diff
@@ -3,10 +3,51 @@
...
+		  </tr>
+		</thead>
+		
+		<tbody>
+			
+		</tbody>
+	</table>
+ 
+	  
+	<script src="app.js"></script>
```
DevinFledermaus: can move `script` tag into the `head` tag

app.ts
ThokozaniNqwili: I made a mistake because of my understanding, I will work on it.

DevinFledermaus (changes requested):
Some comments as initial changes

app.ts
```diff
@@ -0,0 +1,183 @@
+import { ajax, css } from "jquery";
+let firstNumber: number = 0;
+let lastNumber: number = 0;
```
DevinFledermaus: No need to set the variable and the type, either give a type, or set it

app.ts
```diff
@@ -0,0 +1,183 @@
...
+
+
+async function fetchRecordCount(): Promise<number> {
+	try {
+		const recordCount = await fetch(`http://localhost:2050/recordCount`);
+		if (!recordCount.ok) {
+			throw new Error('Failed to fetch record count');
+		}
+		const data = await recordCount.json()
```
```suggestion
		const data = JSON.parse(recordCount)
```

app.ts
```diff
@@ -0,0 +1,183 @@
...
+		return data;
+	} catch (error) {
+		console.error('Error fetching the record count:', error);
+		throw error;
+	}
```
DevinFledermaus, line 17: semicolons please

app.ts
```diff
@@ -0,0 +1,183 @@
...
+
+function fetchColumns(): void {
+	fetch("http://localhost:2050/columns")
+		.then((response: Response) => {
+			return response.json() as Promise<string[]>;
+		})
+		.then((columns: string[]) => {
+			const colArray = columns;
+
+			for (let c = 0; c < colArray.length; c++) {
```
DevinFledermaus: use `for..of` loop please

app.ts
```diff
@@ -0,0 +1,183 @@
...
+	const screenHeight = window.innerHeight;
+	const availableHeight = screenHeight - 105 // subtracts the space used from the screeen
+	let rowHeight = 35
+	if (availableHeight <= 0) {
+		return 0;
+	} else {
+		let maxRows = Math.floor(availableHeight / rowHeight);
+		return maxRows;
+	}
+}
```
DevinFledermaus, line 45: remember semicolons please

app.ts
```diff
@@ -0,0 +1,183 @@
...
+async function displayRecords(): Promise<void> {
+	try {
+		$('#loader').show()
+		let count = await fetchRecordCount() - 1;
+		let calculatedRows = adjustRowsByScreenHeight();
+		const inputValue = $('#searchInput').val() as number;
+		if (calculatedRows === 0) {
+			lastNumber = firstNumber
+		}
+		if (firstNumber < 0) {
```
DevinFledermaus: `else if` and move in the same line as `}`

app.ts
```diff
@@ -0,0 +1,183 @@
...
+			firstNumber = 0;
+			lastNumber = firstNumber + (calculatedRows - 1);
+		} else {
+			lastNumber = firstNumber + (calculatedRows - 1);
+		}
+		let records;
+		if (lastNumber <= count && lastNumber >= 0) {
+			records = await fetchRecords(firstNumber, lastNumber);
```
DevinFledermaus, line 85:
```suggestion
		if (lastNumber <= count && lastNumber >= 0) {
			let records = await fetchRecords(firstNumber, lastNumber);
```

app.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+async function leftArrow(): Promise<void> {
+	$('#searchInput').val('')
+	let count = await fetchRecordCount() - 1;
+	const firstRow = document.querySelector("#recordsTable tbody .row:first-child"); // retrieves the first row
+	if (firstRow) { // checks if the first row exists
+		const cells = firstRow.querySelectorAll("td");
+		const firstRecord: string[] = [];
+		cells.forEach((cell) => {
```
DevinFledermaus: Rather make use of a `for..of` loop, even a for loop is fine, but for of is preferred

app.ts
```diff
@@ -0,0 +1,183 @@
...
+		}
+		const data = await recordCount.json()
+		return data;
+	} catch (error) {
+		console.error('Error fetching the record count:', error);
+		throw error;
+	}
+}
+
+function fetchColumns(): void {
```
```suggestion
fetchColumns(): void {
```

app.ts
```diff
@@ -0,0 +1,183 @@
...
+	fetch("http://localhost:2050/columns")
+		.then((response: Response) => {
+			return response.json() as Promise<string[]>;
+		})
+		.then((columns: string[]) => {
+			const colArray = columns;
```
DevinFledermaus: theres no catch

app.ts
```diff
@@ -0,0 +1,183 @@
...
+	if (!response.ok) {
+		throw new Error("Sorry, there's a problem with the network");
+	}
+	return response.json();
+}
+
+async function displayRecords(): Promise<void> {
+	try {
+		$('#loader').show()
+		let count = await fetchRecordCount() - 1;
```
DevinFledermaus: You shouldn't have to make the backend calls the whole time, call once and save then in variables to be reused

DevinFledermaus (changes requested):
Your backend calls are inconsistent, either stick to `fetch .then .catch`, or `async await`. We tend to make more use of the `fetch .then .catch`. but just keep it consistent, if you do it one way stick, keep it the same

app.ts
```diff
@@ -0,0 +1,218 @@
+import { ajax, css } from "jquery";
+let firstNumber = 0;
+let lastNumber = 0;
+let backend: string = "http://localhost:2050";
```
DevinFledermaus: Try for 2 global variables

app.ts
```diff
@@ -0,0 +1,218 @@
...
+	try {
+		const response = await fetch(`${backend}/recordCount`);
+		if (!response.ok) {
+			throw new Error('Failed to fetch record count');
+		}
+		return response.json()
+	} catch (error) {
+		throw new Error('Error fetching the record count')
+	};
+};
```
DevinFledermaus: No need for `;` after declaring a function

app.ts
```diff
@@ -0,0 +1,218 @@
...
+
+async function fetchColumns(): Promise<string[]> {
+	try {
+		const response = await fetch(`${backend}/columns`);
+		if (!response.ok) {
+			throw new Error('Failed to fetch columns');
+		}
+		const jsonText = await response.text();
+		const columns: string[] = await JSON.parse(jsonText);
```
```suggestion
		const columns: string[] = JSON.parse(response.text());
```

app.ts
```diff
@@ -0,0 +1,218 @@
...
+		const columns = await fetchColumns();
+		for (const col of columns) {
+			$(".head").append(`<th>${col}</th>`);
+		};
+	} catch (error) {
+		throw new Error('Error creating table');
+	};
+};
+
+function adjustRowsByScreenHeight() {
```
DevinFledermaus: no need to use the `function` keyword and provide a return type

app.ts
```diff
@@ -0,0 +1,218 @@
...
+		return response.json();
+	} catch (error) {
+		throw new Error('Error fetching records from server');
+	};
+};
+
+async function displayRecords(): Promise<void> {
+	try {
+		$('#loader').show();
+		let count = await fetchRecordCount() - 1;
```
DevinFledermaus: still not done

app.ts
```diff
@@ -0,0 +1,218 @@
...
+async function fetchRecords(from: number, to: number): Promise<any[]> {
+	try {
+		const response = await fetch(`${backend}/records?from=${from}&to=${to}`);
+		if (!response.ok) {
+			throw new Error("Sorry, there's a problem with the network");
+		}
+		return response.json();
+	} catch (error) {
+		throw new Error('Error fetching records from server');
+	};
```
DevinFledermaus: No need for the async anymore, and can remove the try and catch
```suggestion
	return fetch(`${backend}/records?from=${from}&to=${to}`)
		.then(res => {
			if (!res.ok) {
				throw "Sorry, there's a problem with the network";
			}
			return res.json();
		}).catch(err => {
			throw 'Error fetching records from server ' + err;
		});
```

DevinFledermaus (changes requested):

app.ts
```diff
@@ -0,0 +1,301 @@
+import { ajax, css } from "jquery";
+class GlobalVariables {
```
FritzOnFire: Please leave an empty line between your imports and class definitions.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	firstNumber = 0;
+	lastNumber = 0;
+	backend = "http://localhost:2050";
+	resizeTimeout = 0;
+	//fetches the number of records from localhost
+	fetchRecordCount(): Promise<number> {
```
FritzOnFire: Please leave and empty line between your class attributes and your functions.

app.ts
```diff
@@ -0,0 +1,301 @@
+import { ajax, css } from "jquery";
+class GlobalVariables {
+	firstNumber = 0;
+	lastNumber = 0;
+	backend = "http://localhost:2050";
+	resizeTimeout = 0;
```
FritzOnFire: Please add a type to each of these variables.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	//fetches the number of records from localhost
```
FritzOnFire: Please change this to a javadoc style comment. and instead of saying `localhost`, rather say "the `backend` attribute on this class"

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	fetchRecordCount(): Promise<number> {
+		return fetch(`${this.backend}/recordCount`)
+			.then((res) => {
```
FritzOnFire: Either remove the `()` around the `res`, or specify the correct type for `res`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+				if (!res.ok) {
+					throw 'Failed to fetch record count';
+				}
+				return res.text();
+			})
+			.then((data) => {
+				return JSON.parse(data);
+			})
```
FritzOnFire: This can be simplified by just returning `res.json()`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			.catch((err) => {
+				throw 'Error fetching the record count: ' + err;
+			})
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	}
+	//fetches columns from localhost
+	fetchColumns(): Promise<string[]> {
```
FritzOnFire: Please add an empty line between functions.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+				return res.text();
+			})
+			.then((data) => {
+				return JSON.parse(data);
+			})
+			.catch((err) => {
+				throw 'Error fetching the record count: ' + err;
+			})
+	}
+	//fetches columns from localhost
```
FritzOnFire: Please change this comment to a javadoc style comment. And similar comment as before about localhost.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	fetchColumns(): Promise<string[]> {
+		return fetch(`${this.backend}/columns`)
+			.then(res => {
+				if (!res.ok) {
+					throw 'Failed to fetch columns';
+				}
+				return res.text();
+			})
+			.then((data) => {
+				return JSON.parse(data);
```
FritzOnFire: This can be simplified by just returning `res.json()`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			}).catch(err => {
+				throw 'Error fetching columns' + err;
+			})
```
FritzOnFire, line 23: Missing `;`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	}
+	//fetches records from localhost
+	fetchRecords(from: number, to: number): Promise<any[]> {
```
FritzOnFire: Please leave an empty line between functions.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+				}
+				return res.text();
+			})
+			.then((data) => {
+				return JSON.parse(data);
+			}).catch(err => {
+				throw 'Error fetching columns' + err;
+			})
+	}
+	//fetches records from localhost
```
FritzOnFire: Please change this comment to a javadoc style comment, and similar comment about localhost as before.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	fetchRecords(from: number, to: number): Promise<any[]> {
+		return fetch(`${this.backend}/records?from=${from}&to=${to}`)
+			.then(res => {
+				if (!res.ok) {
+					throw "Sorry, there's a problem with the network";
+				}
+				return res.text();
+			})
+			.then((data) => {
+				return JSON.parse(data);
```
FritzOnFire: This can be simplified by just returning `res.json()`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			}).catch(err => {
+				throw 'Error fetching records from server ' + err;
+			})
```
FritzOnFire, line 38: Missing `;`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	}
+}
+
+const globalVars = new GlobalVariables();
+let recordCount: number;
```
FritzOnFire: Global variables are bad, please remove them.

I noticed that previous comments on the PR indicated a number of global variables to aim for. Please ignore that, as it was meant to be a worst case scenario in case the creator of the PR could not figure out how to do it.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+
+globalVars.fetchRecordCount()
+	.then((count) => {
```
FritzOnFire: Please remove the `()` around your `count` or specify a type.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		// Assign the fetched count to the variable
+		recordCount = count - 1;
+		displayRecords(recordCount);
+	})
+	.catch(err => {
+		throw 'Error fetching record count' + err;
```
FritzOnFire: Nothing seems to be catching this throw. I would recommend that you rather handle the error here, instead of throwing.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	})
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+//Initializes the table head
+function createTable() {
```
FritzOnFire: Please leave an empty line before a function declaration.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+globalVars.fetchRecordCount()
+	.then((count) => {
+		// Assign the fetched count to the variable
+		recordCount = count - 1;
+		displayRecords(recordCount);
+	})
+	.catch(err => {
+		throw 'Error fetching record count' + err;
+	})
+//Initializes the table head
```
FritzOnFire: Please change this comment to a javadoc style comment.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+function createTable() {
+	return globalVars.fetchColumns()
+		.then(columns => {
+			for (const col of columns) {
+				$(".head").append(`<th>${col}</th>`);
+			}
+		})
+		.catch(err => {
+			throw 'Error creating table' + err;
+		})
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+//calculates the number of rows that can fit the screen
+const adjustRowsByScreenHeight = (): number => {
```
FritzOnFire: Please leave an empty line after the previous function.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		.then(columns => {
+			for (const col of columns) {
+				$(".head").append(`<th>${col}</th>`);
+			}
+		})
+		.catch(err => {
+			throw 'Error creating table' + err;
+		})
+}
+//calculates the number of rows that can fit the screen
```
FritzOnFire: Please add a space after the `//`

app.ts
FritzOnFire: This function seems to return something, but the name of the function sounds like it is doing something instead, please rename it to something more fitting.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+const adjustRowsByScreenHeight = (): number => {
+	//retrieves screen height
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	const screenHeight = window.innerHeight;
+	//subtracts the space used from the screeen
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	let rowHeight = 35;
+	if (availableHeight <= 0) {
+		return 0;
+	} else {
+		let maxRows = Math.floor(availableHeight / rowHeight);
+		return maxRows;
+	}
+}
+//calls to re-display records when screen is adjusted
+$(window).on('resize', () => {
```
FritzOnFire: Please leave an empty space between the previous variable and this line.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	const availableHeight = screenHeight - 105;
+	let rowHeight = 35;
+	if (availableHeight <= 0) {
+		return 0;
+	} else {
+		let maxRows = Math.floor(availableHeight / rowHeight);
+		return maxRows;
+	}
+}
+//calls to re-display records when screen is adjusted
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	//subtracts the space used from the screeen
+	const availableHeight = screenHeight - 105;
+	let rowHeight = 35;
+	if (availableHeight <= 0) {
+		return 0;
+	} else {
+		let maxRows = Math.floor(availableHeight / rowHeight);
+		return maxRows;
+	}
+}
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+$(window).on('resize', () => {
+	clearTimeout(globalVars.resizeTimeout);
+	globalVars.resizeTimeout = setTimeout(async () => {
+		await displayRecords(recordCount);
+		// priotizes the search input value if available
+		let inputValue = $('#searchInput').val();
+		if (inputValue !== '') {
+			await updateRecordsAndResize(Number(inputValue));
+		}
+	}, 250)
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+})
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+//display records from first to last number
+async function displayRecords(recordCount: number): Promise<void> {
```
FritzOnFire: Please leave and empty space above your functions.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	globalVars.resizeTimeout = setTimeout(async () => {
+		await displayRecords(recordCount);
+		// priotizes the search input value if available
+		let inputValue = $('#searchInput').val();
+		if (inputValue !== '') {
+			await updateRecordsAndResize(Number(inputValue));
+		}
+	}, 250)
+})
+//display records from first to last number
```
FritzOnFire: Please change this comment to a javadoc style one.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+							//highlights the searched row
+							lastRow.css('background-color', '#DDC0B4');
+						}
+						$("tbody").append(lastRow);
+						$('#loader').hide();
+					}
+				})
+				.catch((err) => {
+					throw 'Error fetching records' + err;
+				})
```
FritzOnFire: There is a lot of code duplication happening here, could you move this to a function and then rather use the function here and in the other branch of the `if`?

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		}
+	} else {
+		throw new Error("Invalid numeric values detected.");
+	}
+}
+
+async function updateRecordsAndResize(inputValue: number) {
+	//check if the search input
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		}
+	} else {
+		throw new Error("Invalid numeric values detected.");
+	}
+}
+
+async function updateRecordsAndResize(inputValue: number) {
+	//check if the search input 
+	if (inputValue < 0 || inputValue > recordCount) {
+		//opens modal if search input is not within range
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		$('.modal').css('display', 'block');
+		$('.content').append(`<p>${inputValue} is not a number within the range. Please try a different number</p>`);
+		$('#page').empty().append(`Showing record: ${globalVars.firstNumber} - ${globalVars.lastNumber}`);
+		//empties search bar
+		$('#searchInput').val('');
+		return
```

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	}
+	let calculatedRows = adjustRowsByScreenHeight();
+	//divides the calculated max rows in half
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+//calls to re-display records when screen is adjusted
+$(window).on('resize', () => {
+	clearTimeout(globalVars.resizeTimeout);
+	globalVars.resizeTimeout = setTimeout(async () => {
+		await displayRecords(recordCount);
+		// priotizes the search input value if available
+		let inputValue = $('#searchInput').val();
+		if (inputValue !== '') {
+			await updateRecordsAndResize(Number(inputValue));
```
FritzOnFire: `updateRecordsAndResize` can throw an error, please catch it and handle it.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	const halfRange = Math.floor(calculatedRows / 2);
+	globalVars.firstNumber = Math.max(0, inputValue - halfRange);
+	globalVars.lastNumber = Math.min(recordCount, globalVars.firstNumber + (calculatedRows - 1));
+	await displayRecords(recordCount);
+}
+
+async function rightArrow(): Promise<void> {
+	$('#page').empty();
+	$('#searchInput').val('');
+	//retrieves the last row
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	const lastRow = document.querySelector("#recordsTable tbody .row:last-child");
+	// checks if the last row exists
+	if (lastRow) {
+		const cells = lastRow.querySelectorAll("td");
+		const lastRecord = [];
+		for (const cell of Array.from(cells)) {
+			lastRecord.push(cell.textContent || "");
+		}
+		//determines te value in the last row
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		const lastID = parseFloat(lastRecord[0]);
+		// checks if the last value is within range
+		if (0 <= lastID && lastID <= (recordCount)) {
+			const tbody = $("tbody");
+			//empties the table
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			tbody.empty();
+			//calculates the first number of the page
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			//display the new records
+			await displayRecords(recordCount);
+		}
+	}
+}
+
+async function leftArrow(): Promise<void> {
+	$('#page').empty();
+	$('#searchInput').val('');
+	//retrieves the first row
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	const firstRow = document.querySelector("#recordsTable tbody .row:first-child");
+	//checks if the first row exists
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	if (firstRow) {
+		const cells = firstRow.querySelectorAll("td");
+		const firstRecord = [];
+		for (const cell of Array.from(cells)) {
+			firstRecord.push(cell.textContent || "");
+		}
+		//determines te value in the first row
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		const firstID = parseFloat(firstRecord[0]);
+		//checks if the first value is within range
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		if (0 <= firstID && firstID <= (recordCount)) {
+			const tbody = $("tbody");
+			//empties the table
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			tbody.empty();
+			const calculatedRows = adjustRowsByScreenHeight();
+			//calculates the last number of the page
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			globalVars.lastNumber = firstID - 1;
+			//uses the last number to calculate first number
```
FritzOnFire: Please add a space after the `//`

app.ts
```diff
@@ -0,0 +1,301 @@
...
+		//calls to calculate the range once button is clicked
+		await updateRecordsAndResize(inputValue);
+	})
+	$('#closeModalBtn').on("click", () => {
+		$('.content').empty();
+		//closes modal
+		$('.modal').css('display', 'none');
+	})
+	$('.arrow-right').on('click', () => {
+		rightArrow();
```
FritzOnFire: This function can throw and error, please catch it and handle it.

app.ts
FritzOnFire: Please add a return type to this function

app.ts
```diff
@@ -0,0 +1,301 @@
...
+			globalVars.lastNumber = firstID - 1;
+			//uses the last number to calculate first number
+			globalVars.firstNumber = globalVars.lastNumber - (calculatedRows - 1);
+			await displayRecords(recordCount);
+		}
+	}
+}
+
+window.onload = () => {
+	createTable();
```
FritzOnFire: This function can throw an error, please catch it and handle it.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	displayRecords(recordCount);
```
FritzOnFire: This function can throw an error, please catch it and handle it.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+				.catch((err) => {
+					throw 'Error fetching records' + err;
+				})
+		}
+	} else {
+		throw new Error("Invalid numeric values detected.");
+	}
+}
+
+async function updateRecordsAndResize(inputValue: number) {
```
FritzOnFire: Please add a return type to this function.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+
+window.onload = () => {
+	createTable();
+	displayRecords(recordCount);
+	$('#btnSearch').on("click", async (event) => {
+		event.preventDefault();
+		let inputValue = $('#searchInput').val() as number;
+		$('#page').empty();
+		//calls to calculate the range once button is clicked
+		await updateRecordsAndResize(inputValue);
```
FritzOnFire: This function can throw an error, please catch it and handle it.

app.ts
```diff
@@ -0,0 +1,301 @@
...
+	$('#closeModalBtn').on("click", () => {
+		$('.content').empty();
+		//closes modal
+		$('.modal').css('display', 'none');
+	})
+	$('.arrow-right').on('click', () => {
+		rightArrow();
+	})
+	$('.arrow-left').on('click', () => {
+		leftArrow();
```
FritzOnFire: This function can throw and error, please catch it and handle it.

style.css
```diff
@@ -0,0 +1,169 @@
...
+	height: 100%;
+	background-color:#f7e7ce;
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	z-index: 9999;
+  }
+  
+.loader-spinner {
+  border: calc(200px / 10) solid #97694F;
```
FritzOnFire: calc is really really really bad, please don't use it.

style.css
```diff
@@ -0,0 +1,169 @@
...
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+	background-color:#f7e7ce;
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	z-index: 9999;
+  }
```
FritzOnFire: Something happend with your formatting here.

style.css
```diff
@@ -0,0 +1,169 @@
...
+  
+.loader-spinner {
+  border: calc(200px / 10) solid #97694F;
+  border-top: 20px solid #4B3832;
+  border-radius: 100%;
+  width: 100px;
+  height: 100px;
+  animation: spin 2s linear infinite;
+  background-color: #f7e7ce;
+  }
```
FritzOnFire: Formatting issue here.

FritzOnFire (changes requested):

app.ts
```diff
@@ -0,0 +1,265 @@
+import { ajax, css } from "jquery";
+
+class Myclass {
```
CelesteNaude: Please move the class into a separate file.

app.ts
```diff
@@ -0,0 +1,265 @@
...
+		$('.modal').css('display', 'block');
+		$('.content').append(`<p>${inputValue} is not a number within the range. Please try a different number</p>`);
+		$('#page').empty().append(`Showing record: ${myClass.firstNumber} - ${myClass.lastNumber}`);
+		$('#searchInput').val('');
+	}
+	let calculatedRows = calculatingRows();
+	/** divides the calculated max rows in half*/
+	const halfRange = Math.floor(calculatedRows / 2);
+	myClass.firstNumber = Math.max(0, inputValue - halfRange);
+	myClass.lastNumber = Math.min(recordCount, myClass.firstNumber + (calculatedRows - 1));
```
CelesteNaude: If I understand correctly, you show the searched record in the middle of the grid and highlight it. In IMQS we show the most relevant record we're searching for at the top of the grid without formatting. But you can still keep the formatting because it's cool.

app.ts
```diff
@@ -0,0 +1,265 @@
...
+	if (lastRow) {
+		const cells = lastRow.querySelectorAll("td");
+		const lastRecord = [];
+		for (const cell of Array.from(cells)) {
+			lastRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the last row */
+		const lastID = parseFloat(lastRecord[0]);
+		/** checks if the last value is within range */
+		if (0 <= lastID && lastID <= (recordCount)) {
```
CelesteNaude: Don't need the () around `recordCount`

app.ts
```diff
@@ -0,0 +1,265 @@
...
+	const firstRow = document.querySelector("#recordsTable tbody .row:first-child");
+	if (firstRow) {
+		const cells = firstRow.querySelectorAll("td");
+		const firstRecord = [];
+		for (const cell of Array.from(cells)) {
+			firstRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the first row */
+		const firstID = parseFloat(firstRecord[0]);
+		if (0 <= firstID && firstID <= (recordCount)) {
```
CelesteNaude: Same as above

CelesteNaude (changes requested):
Please make sure that all events that the user can fire off is debounced, i.e. search, resize, and buttons.

app.ts
```diff
@@ -0,0 +1,230 @@
+import { ajax, css } from "jquery";
+
+const myClass = new Myclass();
```
FritzOnFire: Global variables are bad, please remove them.

FritzOnFire: Also... please... please give it a better name... This name creates the idea that the Dev that made it isn't even aware of what it is used for.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/** Initializes the table head */
+function createTable(): Promise<string[]> {
```
FritzOnFire: I don't think this is the right name for your function. It seems to be adding headings to a dom element with the class `head`. Which does not sound like it is "creating a table".

I also don't know how I feel about returning an array of strings? Or anything for that matter. In my head a `create*` function is either creating something and returning it, or creating something somewhere and returning a reference to it. With the "returning a reference" being completely optional, and I your case I don't think you should be returning anything.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			}
+			return columns;
+		})
+		.catch(err => {
+			throw 'Error creating table' + err;
+		});
+}
+
+/** calculates the number of rows that can fit the screen */
+const calculatingRows = (): number => {
```
FritzOnFire: ... why make a constant variable that points to a function? You can just make a function. Please change this to just be a normal function.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		myClass.resizeTimeout = setTimeout(() => {
+			displayRecords(recordCount)
+				.then(() => {
+					let inputValue = $('#searchInput').val();
+					if (inputValue !== '') {
+						return searchRecordsAndResize(recordCount);
+					}
+				})
+				.catch(err => {
+					throw new Error('Error occured while resizing' + err);
```
FritzOnFire: This is the last line of defense, you have to handle the error.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+				});
+		}, 250);
+	});
+}
+
+/** display records that fit the screen */
+function displayRecords(recordCount: number): Promise<any[]> {
```
FritzOnFire: Using `any` is bad. Please specify something more specific. But maybe the more important question, why are you returning anything? The function name creates the idea that it is going to display records to a user, not that it is going to be returning the records that it is displaying. Looking at the rest of your code I can also see that you are not making use of the records that you are returning. I really think you should not be returning anything.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	$('#loader').show();
+	const inputValue = $('#searchInput').val() as number;
```
FritzOnFire: Please use `<>` to hard cast values. https://github.com/IMQS/www-lib/tree/master/_codestyle#type-casting

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			return records;
+		}).catch(err => {
+			throw new Error('Error displaying records' + err);
+		});
+}
+
+function calculateFirstAndLastNumbers(calculatedRows: number, recordCount: number) {
+	let firstNumber, lastNumber;
+	if (myClass.firstNumber < 0 || myClass.firstNumber > recordCount) {
+		firstNumber = 0;
```
FritzOnFire: I don't think you should take the user back to the start if `firstNumber` is greater than the `recordCount`... I would recommend you set it to the highest value possible.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		.then(records => {
+			$('#page').empty().append(`Showing record: ${firstNumber} - ${lastNumber}`);
+			$('#loader').hide();
+			return records;
+		}).catch(err => {
+			throw new Error('Error displaying records' + err);
+		});
+}
+
+function calculateFirstAndLastNumbers(calculatedRows: number, recordCount: number) {
```
FritzOnFire: Please specify a return type.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+					/** assign each record to their column in a specified row */
+					lastRow.append(`<td>${value}</td>`);
+				}
+				if (record.includes(inputValue)) {
+					/** highlights the searched row */
+					lastRow.css('background-color', '#DDC0B4');
+				}
+				$("tbody").append(lastRow);
+			}
+			return records;
```
FritzOnFire: Why are you returning the records? I would assume that this function does the fetch and display in one go so that the calling function does not have to worry about what records where actually retrieved.

Also based on the rest of your code, you never make use of this variable anyway, and based on the function name it does not make sens to return anything.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	if (lastNumber >= recordCount) {
+		$('.arrow-right').hide();
+	} else {
+		$('.arrow-right').show();
+	}
+}
+
+function fetchAndDisplayRecords(firstNumber: number, lastNumber: number, inputValue: number) {
```
FritzOnFire: Please add a return type to this function.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	return myClass.fetchRecords(firstNumber, lastNumber)
+		.then(records => {
+			const tbody = $("tbody");
+			tbody.empty();
+			for (const record of records) {
+				/** creates row for each record*/
```
FritzOnFire: Not sure why you are using javadoc style commenting here -_-... I would prefer just using `//`, but you can use `/*...*/`, just don't use two `*` at the beginning, only one.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+				$("tbody").append(`<tr class="row"></tr>`);
+				const lastRow = $(".row:last");
+				for (const value of record) {
+					/** assign each record to their column in a specified row */
```
FritzOnFire: Same comment as before about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+					lastRow.append(`<td>${value}</td>`);
+				}
+				if (record.includes(inputValue)) {
+					/** highlights the searched row */
```
FritzOnFire: Same comment as before about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			tbody.empty();
+			for (const record of records) {
+				/** creates row for each record*/
+				$("tbody").append(`<tr class="row"></tr>`);
+				const lastRow = $(".row:last");
+				for (const value of record) {
+					/** assign each record to their column in a specified row */
+					lastRow.append(`<td>${value}</td>`);
+				}
+				if (record.includes(inputValue)) {
```
FritzOnFire: .... no... please just check if the first element in the array matches the input value. Checking the other columns makes no sense...

app.ts
FritzOnFire: Soooooo... I can't find an example that @CelesteNaude is talking about... I can only find two examples where we do it the way that you are doing it. But that has a lot to do with the fact that most of the time we filter instead of search.

(BTW @CelesteNaude, the two places is user-management, where we highlight and scroll to the user you just edited and the grids on the map, when you click on something on the map, and highlight and scroll to the feature you clicked on)

app.ts
```diff
@@ -0,0 +1,230 @@
...
+/** calls to re-display records when screen is adjusted */
+function handleResize(recordCount: number) {
+	$(window).on('resize', () => {
+		clearTimeout(myClass.resizeTimeout);
+		myClass.resizeTimeout = setTimeout(() => {
+			displayRecords(recordCount)
+				.then(() => {
+					let inputValue = $('#searchInput').val();
+					if (inputValue !== '') {
+						return searchRecordsAndResize(recordCount);
```
FritzOnFire: `searchRecordsAndResize` does not return anything. Please remove the `return`

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			return records;
+		})
+		.catch(err => {
+			throw new Error('Error fetching records' + err);
+		});
+}
+
+/** recalculates the record range that includes inputValue */
+function searchRecordsAndResize(recordCount: number) {
+	let inputValue = $('#searchInput').val() as number;
```
FritzOnFire: Please use `<>` when hard casting. https://github.com/IMQS/www-lib/tree/master/_codestyle#type-casting

app.ts
FritzOnFire: `searchRecordsAndResize` also contains a `displayRecords`. Which could lead to a double `displayRecords`. You want to avoid that. Please change your code so that `displayRecords` does not get called twice (or more) from one user action.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	if (inputValue < 0 || inputValue > recordCount) {
+		$('.modal').css('display', 'block');
+		$('.content').append(`<p>${inputValue} is not a number within the range. Please try a different number</p>`);
+		$('#page').empty().append(`Showing record: ${myClass.firstNumber} - ${myClass.lastNumber}`);
+		$('#searchInput').val('');
+		return;
+	}
+	let calculatedRows = calculatingRows();
+	/** divides the calculated max rows in half*/
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	myClass.firstNumber = Math.max(0, inputValue - halfRange);
+	myClass.lastNumber = Math.min(recordCount, myClass.firstNumber + (calculatedRows - 1));
+	displayRecords(recordCount);
+}
+
+/** Navigates to the next set of records */
+function rightArrow(recordCount: number) {
+	$('#page').empty();
+	$('#searchInput').val('');
+	/** retrieves the last row */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	const lastRow = document.querySelector("#recordsTable tbody .row:last-child");
+	/** checks if the last row exists */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	if (lastRow) {
+		const cells = lastRow.querySelectorAll("td");
+		const lastRecord = [];
+		for (const cell of Array.from(cells)) {
+			lastRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the last row */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		const lastID = parseFloat(lastRecord[0]);
+		/** checks if the last value is within range */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		if (0 <= lastID && lastID <= recordCount) {
+			$("tbody").empty();
+			/** calculates the first number of the page */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			myClass.firstNumber = lastID + 1;
+			let calculatedRows = calculatingRows();
+			/** calculates the last number of the page */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			myClass.lastNumber = myClass.firstNumber + (calculatedRows - 1);
+			displayRecords(recordCount);
+		}
+	}
+}
+
+function leftArrow(recordCount: number) {
+	$('#page').empty();
+	$('#searchInput').val('');
+	/** retrieves the first row */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	const firstRow = document.querySelector("#recordsTable tbody .row:first-child");
+	if (firstRow) {
+		const cells = firstRow.querySelectorAll("td");
+		const firstRecord = [];
+		for (const cell of Array.from(cells)) {
+			firstRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the first row */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	const lastRow = document.querySelector("#recordsTable tbody .row:last-child");
+	/** checks if the last row exists */
+	if (lastRow) {
+		const cells = lastRow.querySelectorAll("td");
+		const lastRecord = [];
+		for (const cell of Array.from(cells)) {
+			lastRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the last row */
+		const lastID = parseFloat(lastRecord[0]);
```
FritzOnFire: Instead of looping over everything... can't you rather just check the length of `cells` and then only get the value of the first cell, instead of copying over the value of every cell? Feels like overkill to create a whole new array and populate it if you only want one value.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		/** checks if the last value is within range */
+		if (0 <= lastID && lastID <= recordCount) {
+			$("tbody").empty();
+			/** calculates the first number of the page */
+			myClass.firstNumber = lastID + 1;
+			let calculatedRows = calculatingRows();
+			/** calculates the last number of the page */
+			myClass.lastNumber = myClass.firstNumber + (calculatedRows - 1);
+			displayRecords(recordCount);
```
FritzOnFire: With this function not returning anything, you are going to have to handle the error that `displayRecords` can throw.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		const cells = lastRow.querySelectorAll("td");
+		const lastRecord = [];
+		for (const cell of Array.from(cells)) {
+			lastRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the last row */
+		const lastID = parseFloat(lastRecord[0]);
+		/** checks if the last value is within range */
+		if (0 <= lastID && lastID <= recordCount) {
+			$("tbody").empty();
```
FritzOnFire: `dispalyRecords` calls `fetchAndDisplayRecords` which already cleans out every `tbody` on the page, so no need to also call it here.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+	if (firstRow) {
+		const cells = firstRow.querySelectorAll("td");
+		const firstRecord = [];
+		for (const cell of Array.from(cells)) {
+			firstRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the first row */
+		const firstID = parseFloat(firstRecord[0]);
+		if (0 <= firstID && firstID <= recordCount) {
+			$("tbody").empty();
```
FritzOnFire: `dispalyRecords` calls `fetchAndDisplayRecords` which already cleans out every `tbody` on the page, so no need to also call it here.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		$('.arrow-right').hide();
+	} else {
+		$('.arrow-right').show();
+	}
+}
+
+function fetchAndDisplayRecords(firstNumber: number, lastNumber: number, inputValue: number) {
+	return myClass.fetchRecords(firstNumber, lastNumber)
+		.then(records => {
+			const tbody = $("tbody");
```
FritzOnFire: I don't think it is a good idea to get every `tbody` on the page, rather change this to use an `id`

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			firstRecord.push(cell.textContent || "");
+		}
+		/** determines te value in the first row */
+		const firstID = parseFloat(firstRecord[0]);
+		if (0 <= firstID && firstID <= recordCount) {
+			$("tbody").empty();
+			const calculatedRows = calculatingRows();
+			myClass.lastNumber = firstID - 1;
+			myClass.firstNumber = myClass.lastNumber - (calculatedRows - 1);
+			displayRecords(recordCount);
```
FritzOnFire: With this function not returning anything, you are going to have to handle the error that `displayRecords` can throw.

app.ts
FritzOnFire: The comment on this line is still true.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			});
+			$('.arrow-right').on('click', () => {
+				rightArrow(recordCount);
+			});
+			$('.arrow-left').on('click', () => {
+				leftArrow(recordCount);
+			});
+		})
+		.catch(err => {
+			throw new Error('Error fetching record' + err);
```
FritzOnFire: You are the last line of defense, you have to handle the error here.

FritzOnFire: Also, because you don't have catches on `displayRecords` and `searchRecordsAndResize` this message will be confusing, as the error might be from `displayRecords` or `searchRecordsAndResize` or `fetchRecordCount`, niether of which could have anything to do with fetching record. Specially `fetchRecordCount`... not sure why that was the error message you settled on.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			displayRecords(recordCount);
+		}
+	}
+}
+
+window.onload = () => {
+	createTable();
+	myClass.fetchRecordCount()
+		.then(count => {
+			let recordCount: number = count - 1;
```
FritzOnFire: No need to specify the type here, please remove it.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			displayRecords(recordCount);
+			handleResize(recordCount);
+			searchRecordsAndResize(recordCount);
```
FritzOnFire: Would this not be to early to call this function? Not sure how the code inside would work with the input being empty. And this could also cause a double `displayRecords` which is bad. So at the very least, please ensure that a `displayRecords` does not happen.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+			$('#btnSearch').on("click", (event) => {
+				event.preventDefault();
+				$('#page').empty();
+				searchRecordsAndResize(recordCount);
```
FritzOnFire: You are going to have to catch and handle the error here.

app.ts
```diff
@@ -0,0 +1,230 @@
...
+		$('.content').empty();
+		$('.modal').css('display', 'none');
+	});
+	$('#searchInput').on('keydown', (event) => {
+		if (event.key === 'e' || event.key === 'E') {
+			event.preventDefault();
+		}
+	});
+	$('#searchInput').on('input', () => {
+		const inputValue = $('#searchInput').val() as string;
```
FritzOnFire: Please use `<>` when hard casting. https://github.com/IMQS/www-lib/tree/master/_codestyle#type-casting

index.html
```diff
@@ -1,13 +1,45 @@
 <!DOCTYPE html>
 <html>
+
 <head>
 	<title>JS Onboard Project</title>
 	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+	<link rel="stylesheet" href="style.css">
+	<script src="/myClass.js" ></script>
+	<script src="/app.js" ></script>
```
FritzOnFire: You don't need the `/`

myClass.js
FritzOnFire, line 1: Please don't include new `.js` files in your PR

myClass.js.map
FritzOnFire, line 1: Please don't include new `.js.map` files in your PR

myClass.ts
```diff
@@ -0,0 +1,48 @@
+class Myclass {
```
FritzOnFire: As mentioned, please rename the class

myClass.ts
```diff
@@ -0,0 +1,48 @@
...
+	firstNumber: number = 0;
+	lastNumber: number = 0;
```
FritzOnFire: These two values are not being used in this class... Maybe you forgot to add methods using them? Or maybe they should not be in this class?

myClass.ts
```diff
@@ -0,0 +1,48 @@
...
+	backend: string = "http://localhost:2050";
+	resizeTimeout: number = 0;
```
FritzOnFire: This value is not being used in this class... Maybe you forgot to add methods that uses it? or maybe it should not be in this class?

myClass.ts
```diff
@@ -0,0 +1,48 @@
...
+				}
+				return res.json();
+			})
+			.catch(err => {
+				throw 'Error fetching columns' + err;
+			});
+	}
+
+	/** fetches records from backend */
+	fetchRecords(from: number, to: number): Promise<any[]> {
```
FritzOnFire: Please don't use `any`

myClass.ts
```diff
@@ -0,0 +1,48 @@
+class Myclass {
+	firstNumber: number = 0;
+	lastNumber: number = 0;
+	backend: string = "http://localhost:2050";
+	resizeTimeout: number = 0;
+
+	/** fetches the number of records from backend */
+	fetchRecordCount(): Promise<number> {
```
FritzOnFire: You saaaaay `number`... yet you return `res.json()`... something is not adding up.

FritzOnFire (changes requested):

app.ts
CelesteNaude: This I remember from the advice that was given to me for my own onboarding project, weird.

app.ts
```diff
@@ -0,0 +1,252 @@
...
+		  });
+	  }	  
+
+	/** Calculates the firstNumber and lastNumber */
+	calculateFirstAndLastNumbers() {
+		let rowsPerPage = this.calculatingRows();
+		if (this.firstNumber < 0) {
+			this.firstNumber = 0;
+		} else {
+			this.firstNumber = this.firstNumber;
```
anzelIMQS: Setting `this.firstNumber` equal to `this.firstNumber`? Sjoe 😢 Is the else then even necessary?

app.ts
```diff
@@ -0,0 +1,252 @@
...
+			this.leftArrow();
+		});
+
+		$('#closeModalBtn').on("click", () => {
+			$('.content').empty();
+			$('.modal').css('display', 'none');
+		});
+
+		$('#searchInput').on('keydown', (event) => {
+			if (event.key === 'e' || event.key === 'E') {
```
anzelIMQS: Please rather make use of regex to stop the user from entering anything besides a digit.

app.ts
```diff
@@ -0,0 +1,252 @@
...
+				event.preventDefault();
+			}
+		});
+
+		$('#searchInput').on('input', () => {
+			const inputValue = <string>$('#searchInput').val();
+			if (inputValue.includes('.')) {
+				$('#searchInput').val(inputValue.replace('.', ''));
+			}
```
anzelIMQS: With regex then these checks won't be necessary since you know the user could only type in digits.

app.ts
```diff
@@ -0,0 +1,252 @@
...
+				}
+			})
+			.catch(err => {
+				alert('Error creating table heading');
+				throw err;
+			});
+	}
+
+	/** calculates the number of rows that can fit the screen */
+	calculatingRows(): number {
```
anzelIMQS:
```suggestion
	getNumberOfRows(): number {
```
With a function name like that it is more descriptive of what it does. calculatingRows sounds like a boolean check like isCalculatingRows? Rather give it the verb of what it is going to do. I mean, since this function returns something how about get? How it got it is not maybe important. It calculated it, someone else calculated it? We just want the number of rows.

app.ts
```diff
@@ -0,0 +1,252 @@
...
+				// Get the value of the first cell 
+				const lastID = parseFloat(cells[0].textContent || "");
+				// checks if the last value is within range 
+				if (0 <= lastID && lastID <= this.recordCount) {
+					// calculates the first number of the page 
+					this.firstNumber = lastID + 1;
+					const calculatedRows = this.calculatingRows();
+					// calculates the last number of the page 
+					this.lastNumber = this.firstNumber + (calculatedRows - 1);
+				}
```
anzelIMQS: The way I understand this function, you only need this part of the logic. You don't have to go and find the last child. You know what it is because you set it in `this.lastNumber`, right? Replace `lastID` with `this.lastNumber` and then you can remove the rest and just call `this.updateAndDisplayRecords` afterward.

app.ts
```diff
@@ -0,0 +1,252 @@
...
+		const firstRow = document.querySelector("#recordsTable tbody .row:first-child");
+		if (firstRow) {
+			const cells = firstRow.querySelectorAll("td");
+			if (cells.length > 0) {
+				const firstID = parseFloat(cells[0].textContent || "");
+				if (0 <= firstID && firstID <= this.recordCount) {
+					const calculatedRows = this.calculatingRows();
+					this.lastNumber = firstID - 1;
+					this.firstNumber = this.lastNumber - (calculatedRows - 1);
+				}
```
anzelIMQS: Similar comment to the `rightArrow` function's comment. You don't have to rely on the HTML. You are the point of truth. Also, maybe a more descriptive function name instead of rightArrow and leftArrow? I mean, what does the function do? Navigate back and forth. Verbs instead.

style.css
```diff
@@ -0,0 +1,168 @@
...
+.heading {
+	text-align: center;
+	text-decoration: underline;
+	color: #4B3832;
+}
+
+.showRecords {
+	top: 0.8%;
+	right: 0;
+	padding: 1rem;
```
anzelIMQS: Rather make use of `px` instead of `rem`

app.ts
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	/** fetching records that fit the screen */
+	updateAndDisplayRecords(): Promise<void> {
+		const inputValue = <string>($('#searchInput').val());
+		let errorMessage = '';
+		if (inputValue !== '') {
+		  errorMessage = 'Failed to search, reload the page and search again';
+		  $('#loader').show();
+		  this.searchRecordsAndResize();
```
anzelIMQS: Why not call this outside of `updateAndDisplayRecords`.
Call both `this.searchRecordsAndResize` and then `this.updateAndDisplayRecords` in your search input event?

anzelIMQS (changes requested):

app.ts
```diff
@@ -0,0 +1,214 @@
...
+class RecordManager {
+	firstNumber: number;
+	lastNumber: number;
+	recordCount: number;
+
+	constructor() {
+		this.firstNumber = 0;
+		this.lastNumber = 0;
+		this.recordCount = 0;
+		this.initialize();
```
FritzOnFire: I do not recommend calling functions that can throw errors in a constructor... You could technically still handle the errors, but the rest of your code would think everything went fine, when it very much did not.

Please call this `initialize` after creating a new instance of `RecordManager`

app.ts
```diff
@@ -0,0 +1,214 @@
...
+			.then(count => {
+				this.recordCount = count - 1;
+				this.updateAndDisplayRecords();
+				this.recordEventHandlers();
+				this.handleResize();
+			})
+			.catch(err => {
+				alert('Error fetching and displaying the table records, reload the page');
+				throw err;
+			});
```
FritzOnFire: So... two things.

1. Returning a promise makes catching errors much much easier for the calling function, so i would recommend always doing it. BUT

2. Here you are kicking off two `async` function calls that do not depend on each other, which is beautiful :) This is a great way to speed up an app, but the calling function still needs to be able to handle the error of the function that you are not waiting for. And so now I will be breaking my own rule about not commenting exact code on an onboarding project. This is how you can wait on multiple `async` functions at the same time:

```suggestion
		let p: Promise<void>[] = [];

		p.push(this.createTableHeader());
		let frcp = dataManager.fetchRecordCount()
			.then(count => {
				this.recordCount = count - 1;
				this.updateAndDisplayRecords();
				this.recordEventHandlers();
				this.handleResize();
			})
			.catch(err => {
				alert('Error fetching and displaying the table records, reload the page');
				throw err;
			});
		p.push(frcp);
		
		return Promise.all(p);
```

app.ts
```diff
@@ -0,0 +1,214 @@
...
+		dataManager.fetchRecordCount()
+			.then(count => {
+				this.recordCount = count - 1;
+				this.updateAndDisplayRecords();
+				this.recordEventHandlers();
+				this.handleResize();
+			})
+			.catch(err => {
+				alert('Error fetching and displaying the table records, reload the page');
+				throw err;
```
FritzOnFire: ???

So you handle the error... and then you `throw` it anyway meaning that you have to handle it again later?

Please do not handle the error here, as who ever is calling `initialize` needs to know that we failed to `initialize` and they need to react accordingly.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+	createTableHeader() {
+		return dataManager.fetchColumns()
+			.then(columns => {
+				for (const col of columns) {
+					$(".head").append(`<th>${col}</th>`);
+				}
+			})
+			.catch(err => {
+				alert('Error creating table heading');
+				throw err;
```
FritzOnFire: Similar comment as before about handling and then throwing again.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+		$('#loader').show();
+		this.calculateFirstAndLastNumbers();
+		this.updateArrowVisibility();
+		return this.fetchAndDisplayRecords()
+			.then(() => {
+				$('#loader').hide();
+			})
+			.catch(err => {
+				alert('Error to fetch and display records, reload the page');
+				throw err;
```
FritzOnFire: Similar comment as before about handling and then throwing again.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+						}
+					}
+					$("tbody").append(lastRow);
+				}
+				$('#page').empty().append(`Showing record: ${this.firstNumber} - ${this.lastNumber}`);
+				$('#loader').hide();
+			})
+			.catch(err => {
+				alert('Error while displaying records, reload the page');
+				throw err;
```
FritzOnFire: Similar comment as before about handling and then throwing again.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+				this.handleResize();
+			})
+			.catch(err => {
+				alert('Error fetching and displaying the table records, reload the page');
+				throw err;
+			});
+	}
+
+	/** Initializes the table head */
+	createTableHeader() {
```
FritzOnFire: Missing return type.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+	searchRecordsAndResize() {
+		let inputValue = <number>($('#searchInput').val());
+		if(inputValue >= 0 && inputValue <= this.recordCount){
+			let calculatedRows = this.getNumberOfCalculatingRows();
+		// divides the calculated max rows in half
+		const halfRange = Math.floor(calculatedRows / 2);
+		this.firstNumber = Math.max(0, inputValue - halfRange);
+		this.lastNumber = Math.min(this.recordCount, this.firstNumber + (calculatedRows - 1));
+		}else{
+			alert('Input value must be between 0 and 999999.');
```
FritzOnFire: Please don't hard code the max value.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+	/** recalculates the record range that includes inputValue fromm user */
+	searchRecordsAndResize() {
+		let inputValue = <number>($('#searchInput').val());
+		if(inputValue >= 0 && inputValue <= this.recordCount){
+			let calculatedRows = this.getNumberOfCalculatingRows();
+		// divides the calculated max rows in half
+		const halfRange = Math.floor(calculatedRows / 2);
+		this.firstNumber = Math.max(0, inputValue - halfRange);
+		this.lastNumber = Math.min(this.recordCount, this.firstNumber + (calculatedRows - 1));
+		}else{
```
FritzOnFire: Your formatting seems to have broken on these few lines.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+	navigateToNextPage() {
+		$('#searchInput').val('');
+		if (0 <= this.lastNumber && this.lastNumber <= this.recordCount) {
+			// calculates the first number of the page 
+			this.firstNumber = this.lastNumber + 1;
+			const calculatedRows = this.getNumberOfCalculatingRows();
+			// calculates the last number of the page 
+			this.lastNumber = this.firstNumber + (calculatedRows - 1);
+		}
+		this.updateAndDisplayRecords();
```
FritzOnFire: I would recommend returning this to allow the function calling `navigateToNextPage` to catch and handle the error `updateAndDisplayRecords` is throwing.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	navigateToPreviousPage() {
+		$('#searchInput').val('');
+		if (0 <= this.firstNumber && this.firstNumber <= this.recordCount) {
+			const calculatedRows = this.getNumberOfCalculatingRows();
+			this.lastNumber = this.firstNumber - 1;
+			this.firstNumber = this.lastNumber - (calculatedRows - 1);
+		}
+		this.updateAndDisplayRecords();
```
FritzOnFire: I would recommend returning this to allow the function calling `navigateToPreviousPage` to catch and handle the error `updateAndDisplayRecords` is throwing.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+					this.searchRecordsAndResize();
+				}
+				this.updateAndDisplayRecords()
+					.then(() => {
+						$('#loader').hide();
+					})
+					.catch(err => {
+						alert('Error occurred while resizing, reload the page');
+						$('#loader').hide();
+						throw err;
```
FritzOnFire: You should not be throwing here. You have already handled the error, and there is no one else who can handle errors above you.

app.ts
```diff
@@ -0,0 +1,214 @@
...
+					});
+			}, 250);
+		});
+	}
+
+	recordEventHandlers() {
+		$('#btnSearch').on('click', (event) => {
+			event.preventDefault();
+			this.searchRecordsAndResize();
+			this.updateAndDisplayRecords();
```
FritzOnFire: You need to handle the error being thrown by `updateAndDisplayRecords`

dataManager.ts
```diff
@@ -0,0 +1,51 @@
...
+	static backend: string = "http://localhost:2050";
+
+	/** fetches the number of records from backend */
+	static fetchRecordCount(): Promise<number> {
+		return fetch(`${this.backend}/recordCount`)
+			.then(res => {
+				if (!res.ok) {
+					throw 'Failed to fetch record count';
+				}
+				return res.json();
```
FritzOnFire:
```suggestion
				return res.text();
```

dataManager.ts
```diff
@@ -0,0 +1,51 @@
...
+			})
+			.then(totalRecords => {
+				if(typeof totalRecords !== 'number'){
```
FritzOnFire: Nice touch :) Now I kinda feel bad for asking you to not use `json`.

dataManager.ts
FritzOnFire, line 1: This is a very creative way to hide the fact that you are using a global variable called `backend`. While there is nothing wrong with using `static`, and even making a class like this, as there truly is a time and a place for everything, its just unfortunately not in this on boarding project. Please remove all the static labels in this class.

This is basically as bad as turning off asynchronous requests in javascript for this project (I have only seen it once, and I am unable to find out how to do it, but I know it exists).

index.html
```diff
@@ -1,13 +1,45 @@
...
+	<link rel="stylesheet" href="style.css">
+	<script src="dataManager.js" ></script>
+	<script src="app.js" ></script>
 </head>
 
 <body>
-	<p>Hello</p>
+	<div id="loader">
+		<div class="loader-spinner"></div>
+	  </div>
```
FritzOnFire: Your formatting is a bit off here.

FritzOnFire (changes requested):
Always try and handle errors as late as possible, otherwise your code will not be very portable and reusable.

PR #49, opened 2023-08-14
Onboarding Project: Meezaan

MeezaanD:


app.ts
```diff
@@ -0,0 +1,295 @@
...
+let totalPages: number;
+let currentValueOfFirstRecord: number = 1;
+let currentFirstRecordIndex: number = 1;
+let currentPage: number = 1;
+let recordsPerPage: number;
+let resizeTimeout: number | undefined;
+let searchResultDisplay = false;
+let isSearchMode = false;
+let searchResultIndexes: number[] = [];
+let specialResizeOccurred = false;
```
DevinFledermaus: Try sticking to two global variables. And if you set a variable no need to give the type aswell

app.ts
```diff
@@ -0,0 +1,295 @@
+window.onload = async function () {
+  $("#loader").hide();
+  await fetchTotalRecordCount();
+  updateRecordsPerPage();
+  fetchAndDisplayColumns();
+  handleSpecialResize();
+};
+// Global variables
+const IMQS: string = "http://localhost:2050";
```
DevinFledermaus: camelCase please

app.ts
```diff
@@ -0,0 +1,295 @@
...
+  resizeTimeout = setTimeout(async () => {
+    await updateRecordsPerPage();
+  }, 300);
+});
+
+// Functionality
+async function fetchTotalRecordCount(): Promise<void> {
+  try {
+    let recordCount: number = await (await fetch(`${IMQS}/recordCount`)).json();
+    totalRecordCount = recordCount;
```
```suggestion
    let recordCount: number = await fetch(`${IMQS}/recordCount`);
    totalRecordCount = JSON.parse(recordCount);
```

app.ts
```diff
@@ -0,0 +1,310 @@
+// Global variables
+const IMQS = "http://localhost:2050";
+let totalRecordCount: number;
+let totalPages: number;
+let currentValueOfFirstRecord = 1;
+let currentFirstRecordIndex = 1;
+let currentPage: number;
+let recordsPerPage: number;
+let searchedIndex: number | null = null;
+let resizeTimeout: number;
```
DevinFledermaus: Way too many global variables, try for max two

app.ts
```diff
@@ -0,0 +1,310 @@
...
+
+
+async function totalRecords(): Promise<void> {
+	try {
+		let recordCountResponse = await fetch(`${IMQS}/recordCount`);
+		let recordCountText = await recordCountResponse.text();
```
DevinFledermaus: You don't need to await here

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		let recordCount = parseInt(recordCountText);
+		totalRecordCount = recordCount;
+	} catch (error) {
+		throw (error)
+	};
+};
+async function displayColumns() {
+	try {
+		let columnsResponse = await fetch(`${IMQS}/columns`);
+		let columns = await columnsResponse.json();
```
DevinFledermaus: Don't need to await here

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		let tableHeaderRow = $("#tableHeaderRow");
+		columns.forEach((columnName: string) => {
```
DevinFledermaus: Could you use a `for..of` loop instead

app.ts
```diff
@@ -0,0 +1,310 @@
...
+	try {
+		fromRecord = Math.max(fromRecord, 0);
+		const maximumRecords = Math.min(
+			totalRecordCount - fromRecord,
+			recordsDisplayed
+		);
+		let response = await fetch(
+			`${IMQS}/records?from=${fromRecord}&to=${fromRecord + maximumRecords - 1
+			}`
+		);
```
DevinFledermaus: You making this call multiple times, try having it a function of it's own, similar to totalRecords, and that way displayData wont need to be wont need to be an async function

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		let dataText = await response.text();
```
DevinFledermaus: await not needed

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		let data = JSON.parse(dataText);
+		let tableData = "";
+		if (currentFirstRecordIndex <= 1) {
+			$("#prevPageButton").hide();
+		} else {
+			$("#prevPageButton").show();
+		};
```
DevinFledermaus: no `;` after`if`statements

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		if (searchResultDisplay) {
+			const searchIndex = searchResultIndexes[0];
+			const searchPage = Math.ceil((searchIndex + 1) / recordsPerPage);
+			const newFromRecord = (searchPage - 1) * recordsPerPage;
+			const newToRecord = newFromRecord + recordsPerPage - 1;
+			const actualFromRecord = Math.min(newFromRecord + 1, totalRecordCount);
+			const actualToRecord = Math.min(newToRecord + 1, totalRecordCount);
+			fromRecord = newFromRecord;
+			recordsDisplayed = recordsPerPage;
+			searchResultIndexes.forEach((searchIndex: any) => {
```
DevinFledermaus: `for..of` please

app.ts
```diff
@@ -0,0 +1,310 @@
...
+			}
+			currentValueOfFirstRecord = parseInt(data[0][0]);
+			currentFirstRecordIndex = fromRecord;
+		} else {
+			fromRecord = Math.max(
+				fromRecord + searchIndexOnPage - (recordsDisplayed - 1), 0
+			);
+			isSearchMode = false;
+		};
+		data.forEach((record: any[]) => {
```
DevinFledermaus: same as above

app.ts
```diff
@@ -0,0 +1,310 @@
...
+			tableData += "<tr>";
+			record.forEach((value: string) => {
```
DevinFledermaus: same as above

app.ts
```diff
@@ -0,0 +1,310 @@
...
+				tableData += `<td>${value}</td>`;
+			});
+			tableData += "</tr>";
+		});
+		$("#tableBody").html(tableData);
+	} catch (error) {
+		throw (error)
+	};
+};
```
DevinFledermaus: No need for `;`

app.ts
```diff
@@ -0,0 +1,310 @@
...
+	const estimatedRowHeight = estimatedRowHeightFactor * 50;
+	recordsPerPage = Math.floor(screenHeight / estimatedRowHeight) - 1;
+	recordsPerPage = Math.max(recordsPerPage - 2, 1);
+	$("#loader").show();
+	$("#tableWrapper").hide();
+	if (searchedIndex !== null) {
+		const searchPage = Math.ceil((searchedIndex + 1) / recordsPerPage);
+		const searchPageIndex = searchedIndex % recordsPerPage;
+		searchedIndex = (searchPage - 1) * recordsPerPage + searchPageIndex;
+		await displayData(
```
DevinFledermaus: no await needed

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		} else {
+			$("#prevPageButton").show();
+		};
+		const lastRecordOnPage = currentFirstRecordIndex + recordsPerPage - 1;
+		if ((lastRecordOnPage === totalRecordCount - 1 || searchedIndex !== null) && lastRecordOnPage === totalRecordCount - 1) {
+			$("#nextPageButton").hide();
+		} else {
+			$("#nextPageButton").show();
+		};
+		let isSearchMode = false;
```
DevinFledermaus: already declared as false in the start of the function, and no changes were made to it

app.ts
```diff
@@ -0,0 +1,310 @@
...
+
+
+async function totalRecords(): Promise<void> {
+	try {
+		let recordCountResponse = await fetch(`${IMQS}/recordCount`);
+		let recordCountText = await recordCountResponse.text();
+		let recordCount = parseInt(recordCountText);
+		totalRecordCount = recordCount;
+	} catch (error) {
+		throw (error)
```
DevinFledermaus: `;` please

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		const maximumRecords = Math.min(
+			totalRecordCount - fromRecord,
+			recordsDisplayed
+		);
+		let response = await fetch(
+			`${IMQS}/records?from=${fromRecord}&to=${fromRecord + maximumRecords - 1
+			}`
+		);
+		let dataText = await response.text();
+		let data = JSON.parse(dataText);
```
```suggestion
		let data = JSON.parse(response.text());
```

app.ts
```diff
@@ -0,0 +1,310 @@
...
+let recordsPerPage: number;
+let searchedIndex: number | null = null;
+let resizeTimeout: number;
+
+
+async function totalRecords(): Promise<void> {
+	try {
+		let recordCountResponse = await fetch(`${IMQS}/recordCount`);
+		let recordCountText = await recordCountResponse.text();
+		let recordCount = parseInt(recordCountText);
```
```suggestion
		let recordCount = parseInt(ecordCountResponse.text());
```

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		if ((lastRecordOnPage === totalRecordCount - 1 || searchedIndex !== null) && lastRecordOnPage === totalRecordCount - 1) {
+			$("#nextPageButton").hide();
+		} else {
+			$("#nextPageButton").show();
+		};
+		let isSearchMode = false;
+		let searchResultIndexes: number[] = [];
+		let searchResultDisplay = false;
+		const searchIndexOnPage = searchResultIndexes[0] - fromRecord;
+		if (searchResultDisplay) {
```
DevinFledermaus: When will this ever be true?

app.ts
```diff
@@ -0,0 +1,310 @@
...
+			fromRecord = newFromRecord;
+			recordsDisplayed = recordsPerPage;
+			searchResultIndexes.forEach((searchIndex: any) => {
+				const rowIndex = searchIndex % recordsPerPage;
+				const rowElement = $("#tableBody tr").eq(rowIndex);
+				rowElement.css("background-color", "var(--results-color)");
+			});
+			searchResultDisplay = false;
+		}
+		else if (!isSearchMode) {
```
DevinFledermaus: Will always go in here as you declare it as false right before the if statement. `else` should be on the same line as the `}`

app.ts
```diff
@@ -0,0 +1,310 @@
...
+				tableData += `<td>${value}</td>`;
+			});
+			tableData += "</tr>";
+		});
+		$("#tableBody").html(tableData);
+	} catch (error) {
+		throw (error)
+	};
+};
+async function updatePages(isNavigation = false) {
```
DevinFledermaus: Wont be an async when displayData is changed

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		const lastRecordIndex = totalRecordCount - 1;
+		const searchIndex = Math.min(searchValue, lastRecordIndex);
+		const searchPage = Math.ceil((searchIndex + 1) / recordsPerPage);
+		const fromRecord = (searchPage - 1) * recordsPerPage;
+		const toRecord = Math.min(fromRecord + recordsPerPage - 1, lastRecordIndex);
+		const response = await fetch(
+			`${IMQS}/records?from=${fromRecord}&to=${toRecord}`
+		);
+		const recordsText = await response.text();
+		const records = JSON.parse(recordsText);
```
```suggestion
		const records = JSON.parse(response.text());
```

app.ts
```diff
@@ -0,0 +1,310 @@
...
+			}
+			recordIndex++;
+		}
+		if (foundIndex !== -1) {
+			searchedIndex = foundIndex;
+			const searchPage = Math.ceil((searchedIndex + 1) / recordsPerPage);
+			const searchPageIndex = searchedIndex % recordsPerPage;
+			currentPage = searchPage;
+			currentFirstRecordIndex = (searchPage - 1) * recordsPerPage;
+			await displayData(currentFirstRecordIndex, recordsPerPage);
```
```suggestion
			displayData(currentFirstRecordIndex, recordsPerPage);
```

app.ts
```diff
@@ -0,0 +1,310 @@
...
+		currentFirstRecordIndex = (lastPage - 1) * recordsPerPage;
+	}
+	if (searchedIndex !== null) {
+		const searchPage = Math.ceil((searchedIndex + 1) / recordsPerPage);
+		const searchPageIndex = searchedIndex % recordsPerPage;
+		searchedIndex = (searchPage - 1) * recordsPerPage + searchPageIndex;
+		await searchMethod(currentValueOfFirstRecord);
+	}
+	let fromRecord = currentFirstRecordIndex;
+	await displayData(fromRecord, recordsPerPage);
```
```suggestion
	displayData(fromRecord, recordsPerPage);
```

app.ts
```diff
@@ -0,0 +1,310 @@
...
+async function totalRecords(): Promise<void> {
+	try {
+		let recordCountResponse = await fetch(`${IMQS}/recordCount`);
+		let recordCountText = await recordCountResponse.text();
+		let recordCount = parseInt(recordCountText);
+		totalRecordCount = recordCount;
+	} catch (error) {
+		throw (error)
+	};
+};
```
DevinFledermaus: No `;` after the declaration of functions

DevinFledermaus (changes requested):
Could you also add a space between the functions

app.ts
```diff
@@ -0,0 +1,323 @@
...
+// Fetch the total number of records from the server.
+function totalRecords(): Promise<number> {
+	return fetch(`${state.IMQS}/recordCount`)
+		.then((recordCountResponse) => recordCountResponse.text())
+		.then((recordCountText) => {
+			const recordCount = JSON.parse(recordCountText);
+			return recordCount;
+		}).catch((error) => {
+			throw error;
+		})
```
anzelIMQS:
```suggestion
		});
```
Missing semi-colon

app.ts
```diff
@@ -0,0 +1,323 @@
...
+function fetchRecords(fromRecord: number, toRecord: number): Promise<any> {
+	if (fromRecord > toRecord) {
+		fromRecord = toRecord
+	}
+	return fetch(`${state.IMQS}/records?from=${fromRecord}&to=${toRecord}`)
+		.then((response) => response.text())
+		.then((dataText) => JSON.parse(dataText))
+		.catch((error) => {
+			throw error;
+		})
```
anzelIMQS:
```suggestion
		});
```
Missing semi-colon

app.ts
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// Fetch a range of records from the server based on specified indices.
+function fetchRecords(fromRecord: number, toRecord: number): Promise<any> {
+	if (fromRecord > toRecord) {
+		fromRecord = toRecord
+	}
+	return fetch(`${state.IMQS}/records?from=${fromRecord}&to=${toRecord}`)
+		.then((response) => response.text())
+		.then((dataText) => JSON.parse(dataText))
```
anzelIMQS: You could just say 
```suggestion
		.then((response) => response.json())
```
Instead of first to text and then JSON.parse.
This is applicable to your other fetches as well.
If it is not valid json, the catch will throw your error at least.

app.ts
```diff
@@ -0,0 +1,323 @@
...
+		.then((recordCountText) => {
+			const recordCount = JSON.parse(recordCountText);
+			return recordCount;
+		}).catch((error) => {
+			throw error;
+		})
+}
+
+// Fetch column names and creates them as table headings
+function fetchColumns() {
```
anzelIMQS: Add a return type. You are returning a promise of something or void?

app.ts
```diff
@@ -0,0 +1,323 @@
...
+				const th = document.createElement("th");
+				th.textContent = columnName;
+				tableHeaderRow.append(th);
+			}
+			const recordCount = await totalRecords();
+			totalPages = Math.ceil(recordCount / state.recordsPerPage);
+			updateScreen();
+		}).catch((error) => {
+			throw error;
+		})
```
anzelIMQS:
```suggestion
		});
```
Missing semi-colon

app.ts
```diff
@@ -0,0 +1,323 @@
...
+	return Math.max(recordsPerPage - 2, 1);
+}
+
+// Handles the input for the search
+function inputHandling(): Promise<void> {
+	return totalRecords()
+		.then((totalRecCount) => {
+			const searchInput = document.getElementById("searchInput") as HTMLInputElement;
+			searchInput.max = (totalRecCount - 1).toString();
+		})
```
anzelIMQS:
```suggestion
		});
```
Missing semi-colon

style.css
anzelIMQS, line 1: Some spaces in between your style objects would be nice and more readable.

app.ts
```diff
@@ -0,0 +1,323 @@
...
+					$("#nextPageButton").show();
+					$("#prevPageButton").show();
+					$("#nextPageButton").removeClass("hidden");
+					$("#loader").hide();
+					$("#tableWrapper").show();
+					if (state.currentFirstRecordIndex >= totalRecCount) {
+						$("#nextPageButton").hide();
+					}
+				}, 150);
+			}).catch((error) => {
```
anzelIMQS:
```suggestion
			})
			.catch((error) => {
```

app.ts
```diff
@@ -0,0 +1,323 @@
...
+			$("#tableWrapper").show();
+			$("#prevPageButton").show();
+			if (searchPage === totalPages) {
+				$("#nextPageButton").hide();
+			} else {
+				$("#nextPageButton").show();
+			}
+		}).catch((error) => {
+			throw error;
+		})
```
anzelIMQS:
```suggestion
		});
```
Missing semi-colon

app.ts
```diff
@@ -0,0 +1,323 @@
...
+				throw (error);
+			});
+	}, 150))
+
+	$("#searchInput").on("keydown", function (e) {
+		if (e.key === "e" || e.key === "E" || e.key === "."
+			|| e.key === "+" || e.key === "-") {
+			e.preventDefault();
+		}
+	})
```
anzelIMQS:
```suggestion
	});
```
Missing semi-colon

app.ts
```diff
@@ -0,0 +1,323 @@
...
+			})
+			.then(() => {
+				$("#loader").hide();
+				$("#tableWrapper").show();
+			}).catch((error) => {
+				throw (error);
+			});
+	}, 150))
+
+	$("#searchInput").on("keydown", function (e) {
```
anzelIMQS:
```suggestion
	$("#searchInput").on("keydown", (e) => {
```
We prefer TypeScript arrow functions.

app.ts
```diff
@@ -0,0 +1,323 @@
...
+		}
+		tableData += "</tr>";
+	}
+	$("#tableBody").html(tableData);
+	$("#loader").hide();
+	$("#tableWrapper").show();
+}
+
+// Search for records by a given value and display them on the page.
+async function searchMethod(searchValue: any) {
```
anzelIMQS:
```suggestion
async searchMethod(searchValue: number): Promise<fill in your return type> {
```
Give return types and parameter type.

app.ts
```diff
@@ -0,0 +1,323 @@
...
+	searchValue = Math.min(searchValue, (await totalRecords()) - 1);
+	let targetPage = Math.ceil((searchValue + 1) / state.recordsPerPage);
+	targetPage = Math.min(targetPage, totalPages);
+	const lastRecordIndex = (await totalRecords()) - 1;
```
anzelIMQS: You really shouldn't have to call `totalRecords` multiple times in the same function. Call it once and store the value that the promise returns. Handle it.

app.ts
```diff
@@ -0,0 +1,323 @@
...
+		.then(() => {
+			$("#loader").hide();
+			$("#tableWrapper").show();
+			$("#prevPageButton").show();
+			if (searchPage === totalPages) {
+				$("#nextPageButton").hide();
+			} else {
+				$("#nextPageButton").show();
+			}
+		}).catch((error) => {
```
anzelIMQS:
```suggestion
		})
		.catch((error) => {
```
Put catches on the next line like the suggestion. The same like you did with the thens

app.ts
```diff
@@ -0,0 +1,323 @@
...
+	}, 150))
+
+	$("#searchInput").on("keydown", function (e) {
+		if (e.key === "e" || e.key === "E" || e.key === "."
+			|| e.key === "+" || e.key === "-") {
+			e.preventDefault();
+		}
+	})
+
+	$("#searchForm").submit(async function (e) {
```
anzelIMQS:
```suggestion
	$("#searchForm").submit(async (e) => {
```

app.ts
```diff
@@ -0,0 +1,323 @@
...
+		$("#tableWrapper").hide();
+		$("#loader").show();
+		state.searchedIndex = null;
+		await searchMethod(searchValue);
+		if (state.searchedIndex !== null) {
+			currentPage = Math.ceil((state.searchedIndex + 1) / state.recordsPerPage);
+			state.currentFirstRecordIndex = Math.max(state.searchedIndex - state.recordsPerPage + 1, 0);
+		};
+		$("#loader").hide();
+		$("#tableWrapper").show();
```
anzelIMQS:
```suggestion
		searchMethod(searchValue)
		.then(() => {
				if (state.searchedIndex !== null) {
			currentPage = Math.ceil((state.searchedIndex + 1) / state.recordsPerPage);
			state.currentFirstRecordIndex = Math.max(state.searchedIndex - state.recordsPerPage + 1, 0);
		}
		$("#loader").hide();
		$("#tableWrapper").show();
		});
```
Handle your promise rather. It feels cleaner than `await`. Excuse the indentation in the suggestion.

anzelIMQS (changes requested):
Partial review

app.ts
FritzOnFire: Please ignore, as for global constant values we are supposed to always use all caps.

app.ts
FritzOnFire: @DevinFledermaus you can't parse a promise.

app.ts
FritzOnFire: @DevinFledermaus you can't parse a promise.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+let currentPage: number;
+
+const state = {
+	IMQS: "http://localhost:2050",
+	currentValueOfFirstRecord: 1,
+	currentFirstRecordIndex: 1,
+	recordsPerPage: 16,
+	searchedIndex: null as number | null,
+	isButtonDisabled: false
+}
```
FritzOnFire: Global variables are bad, please remove them.

I noticed that previous comments on the PR indicated a number of global variables to aim for. Please ignore that, as it was meant to be a worst case scenario in case the creator of the PR could not figure out how to do it.

I would recommend moving them to a class along with the methods that are depended on them.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+
+// Fetch the total number of records from the server.
```
FritzOnFire: Please change this comment to match the javadoc style. This is unfortunately not currently in our style guide.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+function totalRecords(): Promise<number> {
+	return fetch(`${state.IMQS}/recordCount`)
+		.then((recordCountResponse) => recordCountResponse.json())
```
FritzOnFire: Please remove the `()` around `recordCountResponse` or specify a type.

Also, you need to check if the response was good, as `.catch` is not going to catch that for you.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		.catch((error) => {
+			throw error;
+		});
+}
+
+// Fetch column names and creates them as table headings
```
FritzOnFire: Similar javadoc style comment.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+function fetchColumns(): Promise<void> {
+	return fetch(`${state.IMQS}/columns`)
+		.then((columnsResponse) => columnsResponse.json())
```
FritzOnFire: Either remove the `()` around `columnsResponse` or specify a type.

Also, you need to check the response was good, as `.catch` is not going to catch that for you.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		.then(async (columns) => {
```
FritzOnFire: Please specify the type for `columns`

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			const tableHeaderRow = $("#tableHeaderRow");
+			for (const columnName of columns) {
+				const th = document.createElement("th");
+				th.textContent = columnName;
+				tableHeaderRow.append(th);
+			}
+			const recordCount = await totalRecords();
+			totalPages = Math.ceil(recordCount / state.recordsPerPage);
```
FritzOnFire: Your `fetchColumns` function should definitely not be doing anything with the total amount of records, or anything that does not have to do with the columns of the table.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			updateScreen();
+		})
+		.catch((error) => {
```
FritzOnFire: Either remove the `()` around `error` or specify a type.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			throw error;
+		});
+}
+
+// Fetch a range of records from the server based on specified indices.
```
FritzOnFire: Similar comment about javadoc style.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+function fetchRecords(fromRecord: number, toRecord: number): Promise<any> {
+	if (fromRecord > toRecord) {
+		fromRecord = toRecord
+	}
```
FritzOnFire: I would prefer if you rather threw an error here, as this function is clearly not being called correctly.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	return fetch(`${state.IMQS}/records?from=${fromRecord}&to=${toRecord}`)
+		.then((response) => response.json())
```
FritzOnFire: Either remove the `()` around `response` or specify a type.

Also, you need to check if the response was good, as `.catch` is not going to catch that.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		.catch((error) => {
+			throw error;
+		});
+}
+
+// Calculate the number of records that can fit on the screen.
```
FritzOnFire: Similar comment about javadoc style.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+function screenCalculations(screenHeight: number): number {
+	const estimatedRowHeightFactor = 1;
+	const estimatedRowHeight = estimatedRowHeightFactor * 50;
+	const recordsPerPage = Math.floor(screenHeight / estimatedRowHeight) - 1;
+	return Math.max(recordsPerPage - 2, 1);
```
FritzOnFire: So, all in all, you subtract 3 from recordsPerPage... that seems quite aggressive. If you are trying to make space for navigation, then I would recommend creating a new variable by subtracting the pixels from `screenHeight` as the function starts. Also put in a comment explaining where the number of pixels are comming from.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	}
+	return fetch(`${state.IMQS}/records?from=${fromRecord}&to=${toRecord}`)
+		.then((response) => response.json())
+		.catch((error) => {
+			throw error;
+		});
+}
+
+// Calculate the number of records that can fit on the screen.
+function screenCalculations(screenHeight: number): number {
```
FritzOnFire: This function name is to vague. Please rename it to something that at least mentions records.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	const estimatedRowHeightFactor = 1;
+	const estimatedRowHeight = estimatedRowHeightFactor * 50;
+	const recordsPerPage = Math.floor(screenHeight / estimatedRowHeight) - 1;
+	return Math.max(recordsPerPage - 2, 1);
+}
+
+// Handles the input for the search
```
FritzOnFire: Similar comment about javadoc style.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+function inputHandling(): Promise<void> {
+	return totalRecords()
+		.then((totalRecCount) => {
```
FritzOnFire: Either remove the `()` around `totalRecCount` or specify a type.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			const searchInput = document.getElementById("searchInput") as HTMLInputElement;
```
FritzOnFire: Please use `<>` for type casting. https://github.com/IMQS/www-lib/tree/master/_codestyle#type-casting

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			searchInput.max = (totalRecCount - 1).toString();
```
FritzOnFire: Nice

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	let timeoutId: any;
+	return function (...args: any) {
+		clearTimeout(timeoutId);
+		timeoutId = setTimeout(() => {
+			func(...args);
+		}, delay);
+	};
+}
+
+// Display records on the page based on the specified range.
```
FritzOnFire: Similar comment about javadoc style.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+function screenCalculations(screenHeight: number): number {
+	const estimatedRowHeightFactor = 1;
+	const estimatedRowHeight = estimatedRowHeightFactor * 50;
+	const recordsPerPage = Math.floor(screenHeight / estimatedRowHeight) - 1;
+	return Math.max(recordsPerPage - 2, 1);
+}
+
+// Handles the input for the search
+function inputHandling(): Promise<void> {
+	return totalRecords()
```
FritzOnFire: For this project, you only have to fetch the record count once. But I can't force you to do that, so it is maybe more like a suggestion.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		}, delay);
+	};
+}
+
+// Display records on the page based on the specified range.
+async function displayData(fromRecord: number, recordsDisplayed: number) {
+	$("#loader").show();
+	$("#tableWrapper").hide();
+	let isSearchMode = state.searchedIndex !== null;
+	const searchIndexOnPage = isSearchMode ? state.searchedIndex! - fromRecord : 0;
```
FritzOnFire: While there is nothing wrong with making use of `!`, I don't think this is the right place for it.

Please rather make use of an `if` statement to get the compiler to stop complaining.

But now that I am looking at the rest of the code. This value is only being used in one place, and in that scenario the value with ALWAYS by `0`... So I don't think you will even need these two lines.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	fromRecord = Math.max(fromRecord, 0);
```
FritzOnFire: Please don't change the value of a parameter.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	const recordCount = await totalRecords();
+	const maximumRecords = Math.min(recordCount - fromRecord, recordsDisplayed);
+	const toRecord = fromRecord + maximumRecords - 1;
```
FritzOnFire: So if `recordCount` is 500 and `fromRecord` is 499 and `recordsDisplayed` is 10, then your `toRecord` will end up being 500, which is an invalid record id, and if it was valid, then it will only fetch one record from the back-end, instead of the 10 that the calling function asked for...  and there are definitely more than 10 records in the back-end.

Please fix your `fromRecord` and `toRecord` calculations.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	const data = await fetchRecords(fromRecord, toRecord);
+	let tableData = "";
+	if (isSearchMode) {
+		if (currentPage * state.recordsPerPage > recordCount) {
+			recordsDisplayed = recordCount - fromRecord;
```
FritzOnFire: Please don't change the value for parameters.

FritzOnFire: I don't agree with changing the amount of records that need to be displayed. The only scenario where something like that would be ok, is if the the amount of records asked to be displayed is larger than the total records in the back-end.

But in that case I would rather throw an error.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		}
+		state.currentValueOfFirstRecord = parseInt(data[0][0]);
```
FritzOnFire: You can't assume that `data` has anything inside of it. It could be an empty array.

Also, if the first value is not equal to the first parameter you fed `fetchRecords`, then something else is seriously wrong.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		state.currentFirstRecordIndex = fromRecord;
+		if (recordsDisplayed >= 16) {
```
FritzOnFire: Remind me to test something on your machine after this review.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			currentPage = totalPages;
+			state.currentFirstRecordIndex = (totalPages - 1) * state.recordsPerPage;
+		}
+	} else {
+		fromRecord = Math.max(
```
FritzOnFire: Please don't change the value of parameters.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			fromRecord + searchIndexOnPage - (recordsDisplayed - 1), 0);
+	}
```
FritzOnFire: What is the point of this `else` if `fromRecord` is no longer being used in this function after this point?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			tableData += `<td>${value}</td>`;
+		}
+		tableData += "</tr>";
+	}
+	$("#tableBody").html(tableData);
+	$("#loader").hide();
+	$("#tableWrapper").show();
+}
+
+// Search for records by a given value and display them on the page.
```
FritzOnFire: Similar comment about javadoc style.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+async function searchMethod(searchValue: any): Promise<void> {
```
FritzOnFire: You are clearly using this value as if it is a number, so please change the type to `number`

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	const totalRecCount = await totalRecords();
+	searchValue = Math.min(searchValue, totalRecCount - 1);
```
FritzOnFire: Please don't change the value of a parameter.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	let targetPage = Math.ceil((searchValue + 1) / state.recordsPerPage);
+	targetPage = Math.min(targetPage, totalPages);
+	const lastRecordIndex = totalRecCount - 1;
+	const searchIndex = Math.min(searchValue, lastRecordIndex);
+	const searchPage = Math.ceil((searchIndex + 1) / state.recordsPerPage);
```
FritzOnFire: In what scenario would this be different than `targetPage`?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	const fromRecord = Math.max(searchIndex - (state.recordsPerPage - 1), 0);
+	const toRecord = Math.min(fromRecord + state.recordsPerPage - 1, lastRecordIndex);
```
FritzOnFire: If `recordsPerPage` represents the amount of records that can be shown on screen at a time (making this assumption, as there is no comment on the `recordsPerPage` variable).

Then all this adding 1 and subtracting 1 makes it feel like it's not really the amount of records that can be shown on a screen... as this code will most of the time make sure it is 2 less than that. This leads me to believe that `recordsPerPage` is maybe something else? Otherwise you would not have these additions and subtractions.

FritzOnFire: It looks like you don't need this `Math.min` as the `Math.min` when you set the value of `searchIndex` ensures that you can never go over the value here. So please remove it.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	return fetchRecords(fromRecord, toRecord)
+		.then((records) => {
```
FritzOnFire: Either remove the `()` around `records` or specify a type.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			let foundIndex = -1;
+			let recordIndex = 0;
+			for (const record of records) {
+				const idValue = parseInt(record[0]);
+				if (idValue === searchValue) {
+					foundIndex = fromRecord + recordIndex;
+					break;
+				}
+				recordIndex++;
+			}
```
FritzOnFire: If you need the index of an array you are looping over then you should rather make use of a c-style for loop, instead of a `for...of` loop.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			if (foundIndex !== -1) {
+				state.searchedIndex = foundIndex;
+				currentPage = searchPage;
+				state.currentFirstRecordIndex = (searchPage - 1) * state.recordsPerPage;
+				return displayData(state.currentFirstRecordIndex, state.recordsPerPage);
```
FritzOnFire: fetching all the records just to confirm if the index you are looking for is in them and then calling a function that is going to again fetch all the records is very expensive. Either pass the records to the next function, or find away to calculate what you need in the next function.

But that is in general. In this scenario, you know, for a fact, that any value between zero, and the record count fetched from the back-end is valid, so you don't need to ask the back-end for anything.

Also, `fetchRecords` would fail if you asked for invalid record ID's , so you for-loop will never hit in the scenario where the ID you are looking for is missing.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+				$("#nextPageButton").show();
+			}
+		})
+		.catch((error) => {
+			throw error;
+		});
+}
+
+
+// Update the screen layout and display records based on the current screen/window size
```
FritzOnFire: Similar comment about javadoc style.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+async function updateScreen(): Promise<void> {
+	const newScreenHeight = window.innerHeight;
+	totalRecords();
```
FritzOnFire: This line seems to do nothing? Can you remove it? or was it needed for something else?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	state.recordsPerPage = screenCalculations(newScreenHeight);
+	$("#loader").show();
+	$("#tableWrapper").hide();
+	if (state.searchedIndex !== null) {
+		const searchPage = Math.ceil((state.searchedIndex + 1) / state.recordsPerPage);
+		const searchPageIndex = state.searchedIndex % state.recordsPerPage;
+		state.searchedIndex = (searchPage - 1) * state.recordsPerPage + searchPageIndex;
```
FritzOnFire: Why would the `searchedIndex` be wrong? Unless I am understanding `searchedIndex` incorrectly, very possible, as there is no comment on that variable.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		displayData(Math.max(state.searchedIndex - state.recordsPerPage + 1, 0), state.recordsPerPage);
+	} else {
+		const previousFirstRecordIndex = state.currentFirstRecordIndex;
+		currentPage = Math.ceil((state.currentFirstRecordIndex + 1) / state.recordsPerPage);
+		displayData(state.currentFirstRecordIndex, state.recordsPerPage);
```
FritzOnFire: I would recommend only calling `displayData` once after this `if` statement, and only use the `if` statement to calculate and store the value of the first parameter of the function.

Also, based on the rest of the code in this function you probably want to `await` or something on this function call, to only execute the rest of this function after this function has finished.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	}
+	$("#loader").hide();
+	$("#tableWrapper").show();
```
FritzOnFire: Does `displayData` not already do this?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	if (currentPage === totalPages) {
+		$("#nextPageButton").hide();
+	} else {
+		$("#nextPageButton").show();
+	}
+	if (state.currentFirstRecordIndex <= 1) {
```
FritzOnFire: But then how do you get to zero?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		$("#prevPageButton").hide();
+	} else {
+		$("#prevPageButton").show();
+	}
+}
+
+window.onload = async function () {
```
FritzOnFire: Please rather use an arrow function.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	$("#loader").hide();
+	updateScreen();
```
FritzOnFire: This function might throw an error, you need to catch it and handle it.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	fetchColumns();
```
FritzOnFire: This function might throw an error, you need to catch it and handle it.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	inputHandling();
```
FritzOnFire: This function might throw an error, you need to catch it and handle it.

app.ts
FritzOnFire: I don't see any reason for this to be an `async` function.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+	$(window).on("resize", debounce(async () => {
```
FritzOnFire: I don't see any reason why this needs to be an `async` function

app.ts
```diff
@@ -0,0 +1,326 @@
...
+					state.currentFirstRecordIndex = (lastPage - 1) * state.recordsPerPage;
+					return displayData(state.currentFirstRecordIndex, state.recordsPerPage);
+				}
+			})
+			.then(() => {
+				$("#loader").hide();
+				$("#tableWrapper").show();
+			})
+			.catch((error) => {
+				throw (error);
```
FritzOnFire: You are the last line of defense, you have the handle the error at this point.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		updateScreen()
+			.then(async () => {
+				const totalRecCount = await totalRecords();
+				if (currentPage * state.recordsPerPage > totalRecCount - 1) {
+					const lastPage = Math.ceil(totalRecCount / state.recordsPerPage);
+					currentPage = lastPage;
+					state.currentFirstRecordIndex = (lastPage - 1) * state.recordsPerPage;
+					return displayData(state.currentFirstRecordIndex, state.recordsPerPage);
+				}
+			})
```
FritzOnFire: If `updateScreen` is updating the screen is not updating the screen correctly, then you should rather go fix `updateScreen` instead of trying to clean up the mess that it is making.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			.then(() => {
+				$("#loader").hide();
+				$("#tableWrapper").show();
+			})
+			.catch((error) => {
+				throw (error);
+			});
+	}, 250))
```

app.ts
```diff
@@ -0,0 +1,326 @@
...
+
+	$("#searchInput").on("keydown", (e) => {
+		if (e.key === "e" || e.key === "E" || e.key === "."
+			|| e.key === "+" || e.key === "-") {
+			e.preventDefault();
+		}
+	});
+
+	$("#searchForm").submit(async (e) => {
```
FritzOnFire: I don't see any reason that this function needs to be `async`.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		state.searchedIndex = null;
+		searchMethod(searchValue)
+			.then(() => {
+				if (state.searchedIndex !== null) {
+					currentPage = Math.ceil((state.searchedIndex + 1) / state.recordsPerPage);
+					state.currentFirstRecordIndex = Math.max(state.searchedIndex - state.recordsPerPage + 1, 0);
+				};
+			})
+		$("#loader").hide();
+		$("#tableWrapper").show();
```
FritzOnFire: Pretty sure you want to move these two lines in the the `.then`.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		$("#tableWrapper").hide();
+		$("#loader").show();
+		state.searchedIndex = null;
+		searchMethod(searchValue)
+			.then(() => {
+				if (state.searchedIndex !== null) {
+					currentPage = Math.ceil((state.searchedIndex + 1) / state.recordsPerPage);
+					state.currentFirstRecordIndex = Math.max(state.searchedIndex - state.recordsPerPage + 1, 0);
+				};
+			})
```
FritzOnFire, line 200: Missing `;`

FritzOnFire: This function can throw an error, you need to catch it and handle it.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		$("#loader").hide();
+		$("#tableWrapper").show();
+	});
+
+	$("#prevPageButton").on("click", async () => {
```
FritzOnFire: I don't see any reason why this function needs to be `async`.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+		if ($("#prevPageButton").hasClass("hidden")) {
+			return;
+		}
+		$("#prevPageButton").addClass("hidden");
+		state.searchedIndex = null;
+		const firstRecordOfCurrentPage = (currentPage - 1) * state.recordsPerPage;
+		if (firstRecordOfCurrentPage <= 16) {
+			currentPage = 1;
+			state.currentFirstRecordIndex = 1;
```
FritzOnFire: I guess I might be wrong here, as I am still not 100% sure what `recordsPerPage` is. But I am getting the feeling that if it was 3 and currentPage was 3, then this code would always make sure that I can not see the data that I am looking for :(.

FritzOnFire: Should this not be setting the value to zero?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			state.currentFirstRecordIndex -= state.recordsPerPage;
+		}
+		const fromRecord = state.currentFirstRecordIndex;
+		$("#nextPageButton").hide();
+		$("#prevPageButton").hide();
+		$("#tableWrapper").hide();
+		$("#loader").show();
+		displayData(fromRecord, state.recordsPerPage)
+			.then(() => {
+				setTimeout(() => {
```
FritzOnFire: You should not have a `setTimeout` here, Unless there is something wrong with `displayData`. Please remove it.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+					$("#nextPageButton").show();
+					$("#prevPageButton").show();
+					$("#prevPageButton").removeClass("hidden");
+					$("#loader").hide();
+					$("#tableWrapper").show();
+					if (state.currentFirstRecordIndex <= 1) {
```
FritzOnFire: So how would I see the first record?

app.ts
```diff
@@ -0,0 +1,326 @@
...
+						$("#prevPageButton").hide();
+					} else if (currentPage < totalPages) {
+						$("#nextPageButton").show();
+					}
+				}, 250);
+			})
+			.catch((error) => {
+				throw error;
```
FritzOnFire: You are the last line of defense, you need to handle the error here.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+					$("#loader").hide();
+					$("#tableWrapper").show();
+					if (state.currentFirstRecordIndex <= 1) {
+						$("#prevPageButton").hide();
+					} else if (currentPage < totalPages) {
+						$("#nextPageButton").show();
+					}
+				}, 250);
+			})
+			.catch((error) => {
```
FritzOnFire: Either remove the `()` around `error` or specify a type.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+				throw error;
+			});
+	});
+
+	$("#nextPageButton").on("click", async () => {
+		if ($("#nextPageButton").hasClass("hidden")) {
```
FritzOnFire: Clever, very cool :)

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			return;
+		}
+		$("#nextPageButton").addClass("hidden");
+		state.searchedIndex = null;
+		const totalRecCount = await totalRecords();
+		const remainingRecords = totalRecCount - (currentPage * state.recordsPerPage);
```
FritzOnFire: Do you really need `remainingRecords`? I think the second part of the `if` pretty much covers it.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+			currentPage = totalPages;
+		}
+		let fromRecord = state.currentFirstRecordIndex;
+		$("#nextPageButton").hide();
+		$("#prevPageButton").hide();
+		$("#tableWrapper").hide();
+		$("#loader").show();
+		displayData(fromRecord, state.recordsPerPage)
+			.then(() => {
+				setTimeout(() => {
```
FritzOnFire: Similar comment as before about the `setTimeout`

app.ts
```diff
@@ -0,0 +1,326 @@
...
+					$("#nextPageButton").removeClass("hidden");
+					$("#loader").hide();
+					$("#tableWrapper").show();
+					if (state.currentFirstRecordIndex >= totalRecCount) {
+						$("#nextPageButton").hide();
+					}
+				}, 250);
+			})
+			.catch((error) => {
+				throw error;
```
FritzOnFire: You are the last line of defense, you need to handle the error here.

app.ts
```diff
@@ -0,0 +1,326 @@
...
+					$("#prevPageButton").show();
+					$("#nextPageButton").removeClass("hidden");
+					$("#loader").hide();
+					$("#tableWrapper").show();
+					if (state.currentFirstRecordIndex >= totalRecCount) {
+						$("#nextPageButton").hide();
+					}
+				}, 250);
+			})
+			.catch((error) => {
```
FritzOnFire: Either remove the `()` around `error` or specify a type.

index.html
```diff
@@ -1,13 +1,41 @@
 <!DOCTYPE html>
 <html>
 <head>
-	<title>JS Onboard Project</title>
-	<script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+  <title>JS Onboard Project</title>
+  <link rel="stylesheet" href="style.css" />
+  <script type="text/javascript" charset="utf-8" src="third_party/jquery-2.0.3.min.js"></script>
+  <script src="/app.js"></script>
```
FritzOnFire: No need for the `/`

index.html
FritzOnFire, line 1: Please stick to tabs for indentation in this file.

index.html
```diff
@@ -1,13 +1,41 @@
...
 </head>
-
 <body>
-	<p>Hello</p>
+  <div class="navigation">
+    <button id="prevPageButton">Previous Page</button>
+    <form id="searchForm">
+      <div class="form-group">
+        <input type="number" id="searchInput" placeholder="Search ID" min="1" required />
```
FritzOnFire: So I can't search for the first record? Bit of an odd choice.

style.css
FritzOnFire, line 1: Please use tabs as indentation.

style.css
```diff
@@ -0,0 +1,184 @@
...
+
+/* Colors */
+:root {
+  --primary-color: white;
+  --secondary-color: #292929;
+  --tertiary-color: black;
+  --shadow-color: #e3e3e3;
+  --selected-color: #0d6efd;
+}
+
```
FritzOnFire: Definitely want to set your variables first in this file. I would recommend moving it up to be the first thing in this file.

style.css
```diff
@@ -0,0 +1,184 @@
...
+  padding: 14px;
+}
+
+tr:hover {
+	cursor: pointer;
+	color: var(--primary-color) !important;
+	background: var(--selected-color);
+	box-shadow: none !important;
+	border: 1px solid var(--secondary-color);
+  }
```
FritzOnFire: Formatting got a bit weird here

style.css
FritzOnFire, line 1: Please ensure an empty line between classes.

FritzOnFire (changes requested):

app.ts
```diff
@@ -0,0 +1,378 @@
+class InitializeApp {
+	IMQS: string = "http://localhost:2050";
+	/** Current value of the first record being displayed */
+	currentValueOfFirstRecord: number = 0;
+	/** Index of the first record currently displayed on the page */
+	currentFirstRecordIndex: number = 0;
+	/** Current page number (changes dynamically)*/
```
```suggestion
	/** Current page number (changes dynamically) */
```

app.ts
```diff
@@ -0,0 +1,378 @@
...
+	currentPage: number = 1;
+	/** Total number of pages available (changes dynamically) */
+	totalPages: number = 1;
+	/** Default number of records to display per page (changes on screen size) */
+	recordsPerPage: number = 16;
+	/** Index of the record being searched for (null if not searching) */
+	searchedIndex: number | null = null;
+	/** Checks if the button is enabled/disabled*/
```
```suggestion
	/** Checks if the button is enabled/disabled */
```

app.ts
```diff
@@ -0,0 +1,378 @@
...
+					resolve();
+				} catch (error) {
+					reject(error);
+				}
+			};
+		});
+		loadApp
+			.catch(() => {
+				throw new Error('Error during window.onload:');
+			});
```
FritzOnFire: Please do this somewhere else... a constructor should not be doing anything that can result in an error being thrown. And it should also definitely not be doing any asynchronous calls.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	/** Fetch the total number of records from the server */
+	totalRecords(): Promise<number> {
+		return fetch(`${this.IMQS}/recordCount`)
+			.then((recordCountResponse) => {
+				if (!recordCountResponse.ok) {
+					throw new Error('Error trying to get recordCount');
+				}
+				return recordCountResponse.json();
```
FritzOnFire: Pretty sure `/recordCount` does not return a json... Pretty sure you want to return `text()` and then parse the text in the next `.then()`

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			});
+	}
+
+	/** Fetch column names and create them as table headings */
+	fetchColumns(): Promise<string[]> {
+		return fetch(`${this.IMQS}/columns`)
+			.then(columnsResponse => {
+				if (!columnsResponse.ok) {
+					throw new Error('Error trying to fetch the columns');
+					return [];
```
FritzOnFire: Please remove this `return`, as you should not have a `return` statement after a `throw`

app.ts
```diff
@@ -0,0 +1,378 @@
...
+				const tableHeaderRow = $("#tableHeaderRow");
+				for (const columnName of columns) {
+					const th = $("<th>").text(columnName);
+					tableHeaderRow.append(th);
+				}
+				return columns;
+			})
+			.catch(error => {
+				throw error;
+				return [];
```
FritzOnFire: Please remove this `return`, as you should not have a `return` statement after a `throw`

app.ts
```diff
@@ -0,0 +1,378 @@
...
+	/** Fetch records within a specified range */
+	fetchRecords(fromRecord: number, toRecord: number): Promise<string[]> {
+		if (fromRecord > toRecord) {
+			return Promise.reject(new Error('Invalid arguments: fromRecord cannot be greater than toRecord'));
+		}
+		return fetch(`${this.IMQS}/records?from=${fromRecord}&to=${toRecord}`)
+			.then(response => {
+				if (!response.ok) {
+					throw new Error('Network response was not ok');
+					return [];
```
FritzOnFire: Please remove this `return`, as you should not have a `return` statement after a `throw`

app.ts
```diff
@@ -0,0 +1,378 @@
...
+				}
+				return response.json();
+			})
+			.catch(() => {
+				return [];
```
FritzOnFire: Please remove this `return`, as you should not have a `return` statement after a `throw`

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			.catch(error => {
+				throw error;
+				return [];
+			});
+	}
+
+	/** Fetch records within a specified range */
+	fetchRecords(fromRecord: number, toRecord: number): Promise<string[]> {
+		if (fromRecord > toRecord) {
+			return Promise.reject(new Error('Invalid arguments: fromRecord cannot be greater than toRecord'));
```
FritzOnFire: Nothing wrong with this line, but I am curious if you can't just `throw` here. Just a comment, you don't have to do anything.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+						tableData += "<tr>";
+						for (const value of record) {
+							tableData += `<td>${value}</td>`;
+						}
+						tableData += "</tr>";
+					}
+				} else {
+					throw new Error('No valid records found');
+				}
+				$("#tableBody").html(tableData);
```
FritzOnFire, line 144: Nice

app.ts
```diff
@@ -0,0 +1,378 @@
...
+				$("#loader").hide();
+				$("#tableWrapper").show();
+			})
+			.catch(error => {
+				throw error;
```
FritzOnFire: Usually when I throw an error in the `.catch` of a promise, I prefer to return that promise to make it easier to catch the error for the calling function.

Just a comment/suggestion. No need to make a code change... all thou, I have not seen the rest of your code, this might be necessary.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	/** Handle the search method */
+	searchMethod(searchValue: number): void {
+		this.totalRecords()
+			.then(totalRecCount => {
+				if (searchValue < 0 || searchValue >= totalRecCount) {
+					window.alert('Record not found on this database');
+					return;
+				}
```
FritzOnFire: I feel that this check would be much easier if you rather used `await` on `totalRecords()`. It would also help clean up the rest of the function a little.

If you are not sure what this should look like using `await`, please call me to your desk.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+							if (idValue === searchValue) {
+								foundIndex = fromRecord + recordIndex;
+								break;
+							}
+						}
+						if (foundIndex !== -1) {
+							this.searchedIndex = foundIndex;
+							this.currentPage = targetPage;
+							this.currentFirstRecordIndex = (targetPage - 1) * this.recordsPerPage;
+							return this.displayData(this.currentFirstRecordIndex, this.recordsPerPage);
```
FritzOnFire: So, `displayData` does not return anything, so please don't put a return in front of it. But you could change `displayData` to return its promise then this line would be correct.

But above all that. It does not seem like you have addressed my previous comment on this code. https://github.com/IMQS/onboard-javascript/pull/49#discussion_r1321648604

You should not be calling `fetchRecords` twice anywhere in your code. And this PR will be stuck until you can resolve that.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+					} else {
+						this.currentPage = Math.ceil((this.searchedIndex + 1) / this.recordsPerPage);
+						this.currentFirstRecordIndex = Math.max(this.searchedIndex - this.recordsPerPage + 1, 0);
+						return this.displayData(this.currentFirstRecordIndex, this.recordsPerPage);
+					}
+				} else {
+					const previousFirstRecordIndex = this.currentFirstRecordIndex;
+					this.currentPage = Math.ceil((this.currentFirstRecordIndex + 1) / this.recordsPerPage);
+					return this.displayData(this.currentFirstRecordIndex, this.recordsPerPage);
+				}
```
FritzOnFire: Similar to a previous comment on a previous review.

You should only call `displayData` once, at the end of this `.then`. The `if` statements should only figureout what value should go in the first parameter of `displayData`. It makes the code a lot easier to read, and there for, to review.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			});
+	}
+
+	/** Update the screen layout and data display */
+	updateScreen(): void {
+		const newScreenHeight = window.innerHeight;
+		this.recordsPerPage = this.windowAdjustments(newScreenHeight);
+		$("#loader").show();
+		$("#tableWrapper").hide();
+		const totalRecCountPromise = this.totalRecords();
```
FritzOnFire: Based on how you use this, would it not be a lot easier to just `await` on `totalRecords`. Keeping a promise around just to `.then` on it twice feels very weird.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+						return this.displayData(this.currentFirstRecordIndex, this.recordsPerPage);
+					}
+				} else {
+					const previousFirstRecordIndex = this.currentFirstRecordIndex;
+					this.currentPage = Math.ceil((this.currentFirstRecordIndex + 1) / this.recordsPerPage);
+					return this.displayData(this.currentFirstRecordIndex, this.recordsPerPage);
+				}
+			})
+			.then(() => {
+				return totalRecCountPromise;
```
FritzOnFire: Please don't re`return` a promise that you have all ready `.then`ed in the same... statement? I don't even know how to phrase this...

Like, a different function can `.then` on a promise, but I am not asking you to do something like that now.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			})
+			.then(totalRecCount => {
+				if (this.currentPage * this.recordsPerPage > totalRecCount - 1) {
+					//  Go to the last page if necessary 
+					const lastPage = Math.ceil(totalRecCount / this.recordsPerPage);
+					this.currentPage = lastPage;
+					this.currentFirstRecordIndex = (lastPage - 1) * this.recordsPerPage;
+					return this.displayData(this.currentFirstRecordIndex, this.recordsPerPage);
```
FritzOnFire: Wait a min... you all ready call `displayData` in a previous `.then` -_-... Why do you need to call it twice?

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			clearTimeout(timeoutId);
+			timeoutId = setTimeout(() => {
+				func(...args);
+			}, delay);
+		};
+	}
+
+	/** Handles the events such as pagination buttons, input handling and search form */
+	eventHandlers(): void {
+		/** Prevent certain keys in the search input */
```
FritzOnFire: Not sure why you are using javadoc here? I would recommend you rather just use `//`, or `/* ... */` if you really want to use `/*`.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+		$("#searchInput").on("keydown", (e) => {
+			if (e.key === "e" || e.key === "E" || e.key === "."
+				|| e.key === "+" || e.key === "-") {
+				e.preventDefault();
+			}
+		});
+
+		/** Handle form submission for searching */
```
FritzOnFire: Similar comment about incorrect use of javadoc.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+		$("#searchForm").submit((e) => {
+			e.preventDefault();
+			const searchInputValue = <string> $("#searchInput").val() ;
+			const searchValue = Number(searchInputValue);
+			$("#tableWrapper").hide();
+			$("#loader").show();
+			this.searchedIndex = null;
+			this.searchMethod(searchValue);
```
FritzOnFire: This function can throw errors. To keep the code clean, I would recommend changing `searchMethod` to return the promise that it is handling, that way you only have to add a `.catch` here, and handle the error in the `.catch`.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			$("#loader").hide();
+			$("#tableWrapper").show();
+		});
+
+		/** Handle previous page button click */
```
FritzOnFire: Similar comment about incorrect use of javadoc

app.ts
```diff
@@ -0,0 +1,378 @@
...
+						this.currentFirstRecordIndex = 0;
+					} else {
+						this.currentPage--;
+						this.currentFirstRecordIndex -= this.recordsPerPage;
+					}
+					fromRecord = this.currentFirstRecordIndex;
+					return this.displayData(fromRecord, this.recordsPerPage);
+				})
+				.catch((error) => {
+					throw error;
```
FritzOnFire: This is the last line of defense, you have to handle the error here.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+						this.currentPage = 1;
+						this.currentFirstRecordIndex = 0;
+					} else {
+						this.currentPage--;
+						this.currentFirstRecordIndex -= this.recordsPerPage;
+					}
+					fromRecord = this.currentFirstRecordIndex;
+					return this.displayData(fromRecord, this.recordsPerPage);
+				})
+				.catch((error) => {
```
FritzOnFire: Either remove the `()` around `error`, or specify the type.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+					throw error;
+				});
+		});
+
+		/** Handle next page button click */
```
FritzOnFire: Similar comment about incorrect usage of javadoc.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+		$("#nextPageButton").on("click", () => {
+			if ($("#nextPageButton").hasClass("hidden")) {
+				return;
+			}
+			if (this.currentPage >= this.totalPages) {
+				const errorMessage = "Already on the last page";
+				window.alert(errorMessage);
+				throw new Error(errorMessage);
```
FritzOnFire: You are the last line of defense, you can't throw an error here. But it looks like you are all ready handling the error with `window.alert`, so I would just replace this line with a return.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+			}
+			$("#nextPageButton").addClass("hidden");
+			this.searchedIndex = null;
+			let fromRecord = this.currentFirstRecordIndex;
+			$("#nextPageButton").hide();
+			$("#prevPageButton").hide();
+			$("#tableWrapper").hide();
+			$("#loader").show();
+			let totalRecCount: number | void;
```
FritzOnFire: This value is not set anywhere... I think you meant to place it in the `.then` that handles the promise coming from `totalRecords`.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+						this.currentPage++;
+						this.currentFirstRecordIndex += this.recordsPerPage;
+					} else {
+						this.currentPage = this.totalPages;
+					}
+					fromRecord = this.currentFirstRecordIndex;
+					return this.displayData(fromRecord, this.recordsPerPage);
+				})
+				.catch((error) => {
+					throw error;
```
FritzOnFire: You are the last line of defense, you need to handle the error here.

app.ts
```diff
@@ -0,0 +1,378 @@
...
+				});
+		});
+	}
+}
+
+/** Initialize the app when the page loads */
+new InitializeApp();
```
FritzOnFire: This feels out of place without the `window.onload = () => { ... };` mentioned in the readme of this project. Not saying your code has to look like that, but I would struggle to have something to comment about if you still used the example that we provided in the readme.

index.html
FritzOnFire: This comment is still relevant.

FritzOnFire (changes requested):

